PR #8, opened 2017-12-20
Us002a load collection exercise data

madeye-matt:
Added POST, PUT and DELETE operations to the collection exercise service

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -14,6 +14,7 @@ import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.CollectionUtils;
 import org.springframework.validation.BindingResult;
+import org.springframework.validation.annotation.Validated;
 import org.springframework.web.bind.annotation.*;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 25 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -141,24 +147,205 @@
...
+   * @throws CTPException on resource not found
+   */
+  @RequestMapping(method = RequestMethod.POST)
+  public ResponseEntity<?> createCollectionExercise(
+          final @Validated(CollectionExerciseDTO.PostValidation.class) @RequestBody CollectionExerciseDTO collex)
+          throws CTPException {
+    log.info("Creating collection exercise");
+    String surveyId = collex.getSurveyId();
+
+    if (StringUtils.isBlank(surveyId)) {
```
benjefferies: Is this covered by the validation in https://github.com/ONSdigital/rm-collectionexercisesvc-api/pull/4/files#diff-896f7d186ede01b90cee7e3076299b31R29

madeye-matt: When you wrote the comment, yes this is covered by the validation so unnecessary.  However the code has evolved since then (survey can be specified either as a UUID or a business friendly ref so this check exists in the newer code and is necessary)

madeye-matt:
@benjefferies  re. Are we implementing the MVP?

Sure.  A single ticket does not a product make.

PR #15, opened 2018-01-09
Changed /collectionexercises/survey/{id} to return full DTO

madeye-matt:
Previously just returned summary with 3 fields


andrewmil:
Apparently I've not got write access here so I only get the grey tick :(

madeye-matt:
Well it let me merge it anyway ;-)

PR #21, opened 2018-01-23
Refactored to use rm-common-service:10.49.7

madeye-matt:


pom.xml
```diff
@@ -63,6 +63,12 @@
       <version>10.50.2</version>
     </dependency>
 
+    <dependency>
```
benjefferies: Not that I'm for heavily depending on a parent pom for dependency versioning etc but it's already defined in the parent https://github.com/ONSdigital/rm-common-config/blob/master/pom.xml#L62

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointUnitTests.java
```diff
@@ -174,10 +174,10 @@ public void findCollectionExercisesForSurvey() throws Exception {
         .andExpect(jsonPath("$[*].id",
             containsInAnyOrder(COLLECTIONEXERCISE_ID1.toString(), COLLECTIONEXERCISE_ID2.toString())))
         .andExpect(jsonPath("$[*].name", containsInAnyOrder(COLLECTIONEXERCISE_NAME, COLLECTIONEXERCISE_NAME)))
-        .andExpect(
-            jsonPath("$[*].scheduledExecutionDateTime", containsInAnyOrder(COLLECTIONEXERCISE_DATE,
-                COLLECTIONEXERCISE_DATE)));
-
+        // MATTTODO fix brittle date handling
```
benjefferies: I wonder if it would make sense to parse the data into the object model to validate?

pom.xml
benjefferies: Agreed offline that we can do this another time as it doesn't have any impact

PR #23, opened 2018-01-30
Use the SURVEY_ID for CI lookup

benjefferies:
Every collection instrument has a survey ID that can be used to look up
collection instrument. This will mean that if there aren't any
classifiers defined the CI will be selected by SURVEY_ID

https://trello.com/c/E7LNNMuG/54-tbl124-required-for-bb-collection-exercise-publish-select-ci-for-survey-as-default


liamtoozer:
Ran through some tests with @VibhorMehra, and happy that it all passes:

- Smoke test from `rm-integration-test` (for regression)
- Manual test for another survey:
    - Created new collection exercise
    - Loaded sample via UI
    - Loaded collection instrument via UI against the collection exercise (using branch from [here](https://github.com/ONSdigital/ras-backstage/pull/40))
    - Ran a POST request against `/collectionexerciseexecute`

jwgwalton:
Nice test refactoring.

PR #28, opened 2018-02-06
Set a whole bunch of other dates from incoming events

madeye-matt:
Additional information has shed light on which events should be matched to which dates on collection exercise. This PR sets the values from the events.  Period start/end just need to be set so currently are set to scheduled start/end date but this may need to change


LJBabbage:
This was folded into release 15 then merged back to master, so this PR is no longer needed

PR #36, opened 2018-02-26
Us047 live state

madeye-matt:
Implemented the transition from READY_FOR_LIVE to LIVE.  The collection exercise service now listens to messages on the collex-event-message-outbound-exchange and when a message with type Elapsed occurs for an event tag of go_live it will send a go_live event to the corresponding collection exercise state machine, hopefully putting the collection exercise into the LIVE state.  If the state is no READY_FOR_LIVE the event will bounce off and currently no other action is taken than an error will be logged (possibly needs more thought from the business)


jcox-dev:
[US047: Create and View 'Live' State of a CE [M]](https://trello.com/c/6htalBsL/37-us047-create-and-view-live-state-of-a-ce-m)

PR #52, opened 2018-04-24
Remove exercises by party id endpoint

tej99:
We should not be using the party data in this service to determine CE association.
Currently backstage is the only service using this endpoint, which is being removed in https://github.com/ONSdigital/ras-backstage/pull/105. I am removing the endpoint to avoid future misuse.

### DIFF
--- a/API.md
+++ b/API.md
@@ -190,59 +190,6 @@ An `HTTP 404 Not Found` status code is returned if the survey with the specified
 ]
 ```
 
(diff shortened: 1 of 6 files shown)


tej99:
@rhiGriff i think it's because removing code and the relevant tests decreases the average lines covered. Doesn't look like the lines codecov is highlighting were being hit in the tests before.

PR #54, opened 2018-04-25
Added endpoint to query quartz for all scheduled jobs

madeye-matt:
This PR is to cover adding an endpoint to the collection exercise service to query the quartz scheduler and return all of the collection exercise events that are actually scheduled (as opposed to the ones that should be scheduled).

There is work afoot to drive the action service from the scheduled events in the collection exercise service.  Once this work is complete, it will be a valuable monitoring and debugging tool to be able to verify precisely what is scheduled.  There is no concrete requirement for this right now but @pricem14pc has confirmed this will be needed in future.


src/main/java/uk/gov/ons/ctp/response/collection/exercise/schedule/SchedulerConfiguration.java
```diff
@@ -80,11 +85,92 @@ public static boolean unscheduleEvent(final Scheduler scheduler, final Event eve
...
+
+            String eventIdStr = jobDataMap.getString(DataKey.id.name());
+            result.setId(UUID.fromString(eventIdStr));
+
+            Trigger trigger = scheduler.getTrigger(TriggerKey.triggerKey(getJobKey(collexId, tagStr)));
+            result.setTimestamp(trigger.getStartTime());
+
+            return result;
+        } catch (SchedulerException e) {
+            log.error("Failed to get event information for job with key {} - {}", jobKey, e.getMessage());
```
benjefferies: Why not just log the stacktrace rather than just the message and should we reraise the exception to return a 500 on the end point if something is wrong with the service?

madeye-matt: - can log the stack trace, I am just aware that most of the overly chatty logging is probably down to me (see state machine failures and collection exercise events). 
- didn't want to re-throw the exception as it makes using the method with streams ugly but can change

PR #60, opened 2018-05-16
add unlink sample endpoint

tej99:
- add api to unlink sample and collection exercise
https://trello.com/c/IQ8agznp/86-us053-int-remove-a-loaded-sample-13

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -465,13 +465,14 @@ public class CollectionExerciseEndpoint {
      *
      * @param collectionExerciseId the collection exercise to unlink from sample
      * @param sampleSummaryId the collection exercise to unlink from collection exercise
+     * @return successful unlink message
      * @throws CTPException            on resource not found
      */
     @RequestMapping(value = "/unlink/{collectionExerciseId}/sample/{sampleSummaryId}", method = RequestMethod.PUT)
     public ResponseEntity<String> unlinkSampleSummary(
             @PathVariable("collectionExerciseId") final UUID collectionExerciseId,
             @PathVariable("sampleSummaryId") final UUID sampleSummaryId) throws CTPException {
-        log.debug("Entering unlinkSampleSummary with collectionExerciseID {} and sampleSummaryId",
+        log.debug("Entering unlinkSampleSummary with collectionExerciseID {} and sampleSummaryId {}",
                 collectionExerciseId, sampleSummaryId);
 
         CollectionExercise collectionExercise = collectionExerciseService.findCollectionExercise(collectionExerciseId);
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -461,6 +461,34 @@ public class CollectionExerciseEndpoint {
     }
 
     /**
+     * PUT request for unlinking sample summary from a collection exercise
+     *
+     * @param collectionExerciseId the collection exercise to unlink from sample
+     * @param sampleSummaryId the collection exercise to unlink from collection exercise
+     * @throws CTPException            on resource not found
+     */
+    @RequestMapping(value = "/unlink/{collectionExerciseId}/sample/{sampleSummaryId}", method = RequestMethod.PUT)
+    public ResponseEntity<String> unlinkSampleSummary(
+            @PathVariable("collectionExerciseId") final UUID collectionExerciseId,
+            @PathVariable("sampleSummaryId") final UUID sampleSummaryId) throws CTPException {
+        log.debug("Entering unlinkSampleSummary with collectionExerciseID {} and sampleSummaryId",
+                collectionExerciseId, sampleSummaryId);
+
+        CollectionExercise collectionExercise = collectionExerciseService.findCollectionExercise(collectionExerciseId);
+        if (collectionExercise == null) {
+            throw new CTPException(CTPException.Fault.RESOURCE_NOT_FOUND,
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -460,6 +460,35 @@ private void validateConstraints(final CollectionExerciseDTO collexDto) throws C
...
 
+    /**
+     * PUT request for unlinking sample summary from a collection exercise
+     *
+     * @param collectionExerciseId the collection exercise to unlink from sample
+     * @param sampleSummaryId the collection exercise to unlink from collection exercise
+     * @return successful unlink message
+     * @throws CTPException            on resource not found
+     */
+    @RequestMapping(value = "/unlink/{collectionExerciseId}/sample/{sampleSummaryId}", method = RequestMethod.PUT)
```
benjefferies: Maybe `RequestMethod.DELETE` would be more appropriate

src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -460,6 +460,35 @@ private void validateConstraints(final CollectionExerciseDTO collexDto) throws C
...
+        if (collectionExercise == null) {
+            throw new CTPException(CTPException.Fault.RESOURCE_NOT_FOUND,
+                    String.format("%s %s", RETURN_COLLECTIONEXERCISENOTFOUND, collectionExerciseId));
+        }
+
+        collectionExerciseService.removeSampleSummaryLink(sampleSummaryId, collectionExerciseId);
+
+        String responseMessage = "Successfully unlinked collection exercise "
+                + collectionExerciseId + " and sample " + sampleSummaryId;
+        return ResponseEntity.ok(responseMessage);
```
benjefferies: I reckon we should get out of the habit of returning unstructured text from services. We should just return 204

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -165,6 +165,22 @@ public CollectionExercise findCollectionExercise(String surveyRef, String exerci
...
+     * @param collectionExerciseId a collection exercise uuid
+     * @throws CTPException thrown if transition fails
+     */
+    @Override
+    @Transactional
+    public void removeSampleSummaryLink(final UUID sampleSummaryId, final UUID collectionExerciseId)
+            throws CTPException {
+        sampleLinkRepository.deleteBySampleSummaryIdAndCollectionExerciseId(sampleSummaryId, collectionExerciseId);
+
+        transitionScheduleCollectionExerciseToReadyToReview(collectionExerciseId);
```
benjefferies: Although it does the right thing in transitioning the state it's a bit weird as it will transition the state based on a CI_ADDED event which is confusing. For constancy we probably need to add another event and do the transition based on that event

tej99: The events used are CI_SAMPLE_ADDED and CI_SAMPLE_DELETED. But i've changed the logic now to prevent transition getting triggered with a CI_SAMPLE_ADDED event when removing a sample (which would happen if we have multiple sample links for that collection exercise).

src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -460,6 +460,33 @@ private void validateConstraints(final CollectionExerciseDTO collexDto) throws C
 
     }
 
+    /**
+     * PUT request for unlinking sample summary from a collection exercise
```
benjefferies: Maybe just remove the PUT and change to for unlinking sample summary from a collection exercise

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointUnitTests.java
```diff
@@ -303,6 +306,49 @@ public void linkSampleUnitsPut() throws Exception {
         .andExpect(jsonPath("$.sampleSummaryIds[*]", containsInAnyOrder(SAMPLE_SUMMARY_ID1.toString(), SAMPLE_SUMMARY_ID2.toString())));
   }
 
+  /**
+   * Tests put request to unlink sample.
+   *
+   * @throws Exception exception thrown
+   */
+  @Test
+  public void testUnlinkSampleUnitsPut() throws Exception {
```
benjefferies: Maybe remove the Put from the method

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointUnitTests.java
```diff
@@ -303,6 +306,49 @@ public void linkSampleUnitsPut() throws Exception {
...
+
+    actions.andExpect(status().isNoContent())
+            .andExpect(handler().handlerType(CollectionExerciseEndpoint.class))
+            .andExpect(handler().methodName("unlinkSampleSummary"));
+
+    verify(collectionExerciseService, times(1)).removeSampleSummaryLink(SAMPLE_SUMMARY_ID1, COLLECTIONEXERCISE_ID1);
+  }
+
+  /**
+   * Tests put request to unlink sample with unknown collection exercise
```
benjefferies: Maybe remove the Put from the method

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointUnitTests.java
rjweeks: I agree that removing the 'Put' is better naming style.

PR #67, opened 2018-06-11
Added an integration test to test receiving sample upload finished messages

madeye-matt:
# Motivation and Context
This is the second attempt to introduce integration tests to the collection exercise service.  The original attempt failed as it used a simple Java client to send rabbit messages.   This did not work as expected on cloud foundry so had to be backed out.  This version uses spring integration to drive the messaging so should configure ok on cloud foundry (same approach works on other services).

The integration tests that have been added have been designed to play nice with the standard docker setup so there should be no port clashes (i.e. the integration tests can be run at the same time as the full docker env)

This reverts commit 55cbda5b2fae24fa9a08b7ab602c5cd0641914e4.

# What has changed
- when a sample link is activated, a message is sent to collection-outbound-exchange
- an integration test has been added to test sample link activation

# How to test?
- run 'make up' to spin up all docker services
- run 'mvn clean install' and check for successful completion

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointIT.java
+++ b/src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointIT.java
@@ -1,0 +1,161 @@
+package uk.gov.ons.ctp.response.collection.exercise.endpoint;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.Pair;
+import org.junit.Before;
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointIT.java
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * A class to contain integration tests for the collection exercise service
+ */
+@Slf4j
+@RunWith(SpringRunner.class)
+@ContextConfiguration
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+public class CollectionExerciseEndpointIT {
+
+    // TODO pull these from config
```
benjefferies: Is this comment needed?

PR #68, opened 2018-06-11
Changes to allow DB configuration and scaling

mfcrocker:
# Motivation and Context
It turns out that the database parameters in application-cloud.yml are overridden by CloudFoundry. This changes that!

# What has changed
`DataSourceConfiguration` has been added which forces Spring Boot to use the config rather than CloudFoundry's one from buildpacks

# How to test?
You'll need to push to CloudFoundry to see the impact of these changes. The easiest way is to set `spring.datasource.tomcat.max-idle` to a higher number than `max-active`; this will cause Spring Boot to log out a message just after the service starts that max-idle is higher than max-active and give you the figures in the .yml. If it says it's setting it to 4, it's broken (these are the defaults CF was injecting)


mfcrocker:
[Optimise DB Connection Settings](https://trello.com/c/SwXkyTRo/44-optimise-db-connection-settings)

src/main/java/uk/gov/ons/ctp/response/collection/exercise/CollectionExerciseApplication.java
```diff
@@ -1,6 +1,8 @@
 package uk.gov.ons.ctp.response.collection.exercise;
 
 import net.sourceforge.cobertura.CoverageIgnore;
+
+import org.apache.tomcat.jdbc.pool.DataSource;
```
tej99: Unused import

src/main/resources/application-cloud.yml
```diff
@@ -60,6 +60,10 @@ spring:
       min-idle: 3
     continue-on-error: true
     schema: classpath:/schema-cloud.sql
+    user: ${cloud.services.postgresql.connection.username}
+    password: ${cloud.services.postgresql.connection.password}
+    default-schema: sample
```
gemmai95: This should be collection exercise.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/config/DataSourceConfiguration.java
```diff
@@ -0,0 +1,33 @@
...
+        return new CloudFactory().getCloud();
+    }
+    
+    @Bean
+    @ConfigurationProperties(prefix="spring.datasource.tomcat")
+    public DataSource dataSource() {
+        return cloud().getSingletonServiceConnector(DataSource.class, null);
+    }
+
+}
```
gemmai95: 🎨 Missing new line

mfcrocker: Java doesn't require one

src/main/resources/application-cloud.yml
VibhorMehra: Testing this branch with `default-schema: collectionexercise` on prunes.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/config/DataSourceConfiguration.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.ons.ctp.response.collection.exercise.config;
+
+import org.apache.tomcat.jdbc.pool.DataSource;
```
VibhorMehra: This is the wrong data source, it should be 

```java
import javax.sql.DataSource;``

PR #73, opened 2018-06-14
When creating collection exercise create default action plans and casetypedefaults if not exists

andrewmil:
# Motivation and Context
When a collection exercise is created, if the default action plans and case type default do not exist for the given survey, then create them

# What has changed
Add sequence for casetypedefault table
Create new action plans and casetypedefaults if they do not exist for survey at collection exercise execution

# How to test?
Create the first collection exercise for a survey, check that the correct action plans and case type defaults are created
Acceptance tests/Integration test should pass

# Links
[Trello](https://trello.com/c/rnSOtuA1/66-create-case-type-default-in-collection-exercise-for-new-survey)
[rasrm-acceptance-tests PR](https://github.com/ONSdigital/rasrm-acceptance-tests/pull/140)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -685,6 +685,19 @@ public class CollectionExerciseEndpoint {
     }
 
     /**
+     * Get's the list of all scheduled events from quartz (i.e. it's not the list of events as stored in the database,
+     * it's the actual jobs scheduled in quartz)
+     * @return the list of events derived from quartz jobs
+     * @throws SchedulerException thrown if issues getting data from quartz
+     */
+    @RequestMapping(value = "/events/scheduled", method = RequestMethod.GET)
+    public ResponseEntity<List<EventDTO>> getAllScheduledEvents() throws SchedulerException {
+        List<EventDTO> scheduledEvents = SchedulerConfiguration.getAllScheduledEvents(this.scheduler);
+
+        return ResponseEntity.ok(scheduledEvents);
+    }
+
+    /**
      * Endpoint to get a collection exercise by exercise ref and survey ref
      *
      * @param exerciseRef the exercise ref
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -29,6 +29,7 @@ import uk.gov.ons.ctp.response.collection.exercise.representation.CollectionExer
 import uk.gov.ons.ctp.response.collection.exercise.representation.EventDTO;
 import uk.gov.ons.ctp.response.collection.exercise.representation.LinkSampleSummaryDTO;
 import uk.gov.ons.ctp.response.collection.exercise.representation.LinkedSampleSummariesDTO;
+import uk.gov.ons.ctp.response.collection.exercise.representation.SampleLinkDTO;
 import uk.gov.ons.ctp.response.collection.exercise.representation.SampleUnitValidationErrorDTO;
 import uk.gov.ons.ctp.response.collection.exercise.schedule.SchedulerConfiguration;
 import uk.gov.ons.ctp.response.collection.exercise.service.CollectionExerciseService;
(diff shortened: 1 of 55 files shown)


andrewmil:
[Create case type default in collection exercise for new survey.](https://trello.com/c/rnSOtuA1/66-create-case-type-default-in-collection-exercise-for-new-survey)

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -270,21 +270,72 @@ private CollectionExercise newCollectionExerciseFromDTO(CollectionExerciseDTO co
...
+                actionPlan.getId(), survey.getId(), sampleUnitType);
+    }
+
+    /**
+     * Create case type default for action plan, survey and sample unit type if not already exists
+     * @param survey DTO Representation of survey
+     * @param sampleUnitType Sample Unit Type
+     * @param actionPlan DTO Representation of action plan
+     */
+    private void createCaseTypeDefault(SurveyDTO survey, String sampleUnitType, ActionPlanDTO actionPlan) {
```
mfcrocker: If this must logically happen on every call to `createDefaultActionPlan`, I don't think it makes sense to have this as a separate method. Nesting method calls like this hurts readability

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImplTest.java
```diff
@@ -263,23 +269,55 @@ public void testCreateCollectionExerciseCreatesTheActionPlans() throws Exception
...
-  public void testCreateCollectionExerciseExistingActionPlans() throws Exception {
+  public void testCreateCollectionExerciseExistingDefaultActionPlans() throws Exception {
     // Given
     CollectionExerciseDTO toCreate = FixtureHelper.loadClassFixtures(CollectionExerciseDTO[].class).get(0);
     CollectionExercise collectionExercise = FixtureHelper.loadClassFixtures(CollectionExercise[].class).get(0);
     collectionExercise.setExerciseRef(toCreate.getExerciseRef());
     when(collexRepo.saveAndFlush(any())).thenReturn(collectionExercise);
     SurveyDTO survey = FixtureHelper.loadClassFixtures(SurveyDTO[].class).get(0);
     when(this.surveyService.findSurvey(UUID.fromString(toCreate.getSurveyId()))).thenReturn(survey);
+    ActionPlanDTO actionPlanDTO = new ActionPlanDTO();
```
mfcrocker: :art: Would consider putting a couple of blank lines in this and the next test (probably after the when statements) to make it more readable, but maybe that's just me reading it in GitHub rather than an IDE.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -270,21 +270,72 @@ private CollectionExercise newCollectionExerciseFromDTO(CollectionExerciseDTO co
...
+        if (existingCaseTypeDefault != null) {
+            log.debug("Default action plan already exists, SurveyId: {} SampleUnitType: {}",
+                    survey.getId(), sampleUnitType);
+            return;
+        }
+
+        // Create new action plan and associated casetypedefault
+        String shortName = survey.getShortName();
+        String name = String.format("%s %s", shortName, sampleUnitType);
+        String description = String.format("%s %s Case", shortName, sampleUnitType);
```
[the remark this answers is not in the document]
andrewmil: I think the reason for doing this is because we have to pass the `name` and `description` into the `createActionPlan` function and it could be a bit messy to do that inline when calling the function

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
andrewmil: Will look into this. If there's no other situation we're going to want to `createCaseTypeDefaults(/Overrides)` that probably makes sense

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImplTest.java
andrewmil: :+1: Will add a couple more comments to to make it clearer what we're mocking

src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -399,6 +399,7 @@ private void validateConstraints(final CollectionExerciseDTO collexDto) throws C
                             collex.getExerciseRef(), surveyRef));
         }
 
+        // Create new collection exercise, including required action plans and casetypedefaults/overrides
```
tomphp: This comment seems unncessary

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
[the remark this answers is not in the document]
tomphp: I disagree as it add description, however, this is completely stylistic choice.

PR #85, opened 2018-07-12
Decouple formtype social

damien-lloyd:
# Motivation and Context
Formtypes don't exist in social samples so we're dealing with them in the collection exercise service on the basis that they can be null.

# What has changed
Updated tests to create a non-null formtype if the sample is a business; otherwise leave blank for socials. 
Created a delta to allow a nullable formtype

# How to test?
Run all unit tests
Run all integration tests
Build the service
Run all acceptance tests


damien-lloyd:
[Decouple social sample units from 'formtype' in Collection Exercise Service (13)](https://trello.com/c/ECP2gzzV/179-decouple-social-sample-units-from-formtype-in-collection-exercise-service-13)

PR #87, opened 2018-07-19
Remove bi/always create b case at execution

gemmai95:
# Motivation and Context

This change is needed to always create B cases when collection exercises are executed. 

# What has changed

Now the sample unit is always published to case to create a B case.

# How to test?

Tested with case and case-api changes execute a second collection exercise and check B cases are being created as well as BI cases when businesses have respondents enrolled for the survey.

# Links

[Trello] https://trello.com/c/ZWg5l0Kt

Linked PRs:
[Case API] https://github.com/ONSdigital/rm-casesvc-api/pull/24
[Case Service]  https://github.com/ONSdigital/rm-case-service/pull/56


pom.xml
```diff
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>uk.gov.ons.ctp.product</groupId>
             <artifactId>casesvc-api</artifactId>
-            <version>10.49.17</version>
+            <version>10.49.23</version>
```
shahi645: Is there a reason this doesn't match the version that Case SVC is using?

gemmai95: When my PR goes in on casesvc-api I will update this. It is dependent on that one.

PR #88, opened 2018-07-20
Handle survey service client correctly

damien-lloyd:
# Motivation and Context
An issue was raised that we were treating 4xx and 5xx errors alike. A clause was added to take care of this.

# What has changed
Added more exception handling to ensure if there's a 4xx we treat it as such, and if there's a 5xx we also treat it as such.
Wrote a unit test which covers the clause, should work for both 4xx and 5xx 

# How to test?
Run unit test in SurveySvcRestClientImplTest


HacktechLTD:
@madeye-matt those places where it raises a NOT_FOUND if null is returned are still perfectly valid - the logic is the same for resource not found (i.e. return null). This Damien did this PR off the back of an issue I raised in Github, where we would be treating things that weren't 404 as if they were... we wouldn't want to attempt to create something because we believed it was missing.

I think @LJBabbage is gonna have chat with all the teams about what we're doing in Improvements, cos I didn't mean for anybody to pick the issue up and fix it other than our team!

Apols for any confusion/noise generated.

(Nick Grant)

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
```diff
@@ -123,46 +124,55 @@ private SurveyDTO getSurveyDtoFromResponseEntity(ResponseEntity<String> response
...
           surveyId,
           uriComponents.toUri());
-      ResponseEntity<String> responseEntity =
+      responseEntity =
           restTemplate.exchange(uriComponents.toUri(), HttpMethod.GET, httpEntity, String.class);
-
       return getSurveyDtoFromResponseEntity(responseEntity);
     } catch (RestClientException e) {
-      return null;
+      if (responseEntity != null && responseEntity.getStatusCode().is4xxClientError()) {
```
benjefferies: Not sure if this is adding any value. The behaviour spring mvc has is that it will already do this.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
```diff
@@ -123,46 +124,55 @@ private SurveyDTO getSurveyDtoFromResponseEntity(ResponseEntity<String> response
...
-          surveyRef,
-          uriComponents.toUri());
-      ResponseEntity<String> responseEntity =
+      responseEntity =
           restTemplate.exchange(uriComponents.toUri(), HttpMethod.GET, httpEntity, String.class);
 
       return getSurveyDtoFromResponseEntity(responseEntity);
     } catch (RestClientException e) {
-      return null;
+      if (responseEntity != null && responseEntity.getStatusCode().is4xxClientError()) {
```
benjefferies: Not sure if this is adding any value. The behaviour spring mvc has is that it will already do this.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnits.java
```diff
@@ -189,6 +189,8 @@ private void generateSampleUnits(
               sampleUnitGroup.getSampleUnitGroupPK(),
               ex.getMessage());
           log.error("Stack trace: " + ex);
+        } catch (CTPException e) {
+          e.printStackTrace();
```
benjefferies: We shouldn't be printing the stacktrace like this

src/main/resources/database/changelog-master.yml
```diff
@@ -33,3 +33,6 @@ databaseChangeLog:
 
   - include:
       file: database/changes/release-10.49.15/changelog.yml
+
+  - include:
```
benjefferies: This change seems unrelated

src/main/resources/database/changes/release-10.49.16/changelog.yml
```diff
@@ -0,0 +1,11 @@
+databaseChangeLog:
```
benjefferies: This change seems unrelated

src/main/resources/database/changes/release-10.49.16/make_formtype_nullable.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE collectionexercise.sampleunitgroup ALTER COLUMN formtype DROP NOT NULL;
```
benjefferies: This change seems unrelated

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpointIT.java
```diff
@@ -257,8 +260,12 @@ private SampleUnitParent ensureSampleUnitIdIsPropagatedHere(String type) throws
     sampleUnit.setId(id.toString());
     sampleUnit.setSampleUnitRef("LMS0001");
     sampleUnit.setCollectionExerciseId(collex.getId().toString());
-    sampleUnit.setFormType("");
     sampleUnit.setSampleUnitType(type);
+
+    if (type.equalsIgnoreCase("B") || type.equalsIgnoreCase("BI")) {
```
benjefferies: This change seems unrelated

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
HacktechLTD: RestTemplate is not Spring MVC and the previous behaviour was to catch the exception and return null on any RestClientException. 404 is an expected exception where we should return null, but what about any other kind of problems with the REST endpoint we're calling?

It looks to me like we'll lose all error logging and ability to use retry annotations if we just return null whenever there's a problem - we're introducing hard to diagnose flakiness.

benjefferies: Sorry meant MVC will catch the RestClient exception with the generic exception handler

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
```diff
@@ -123,46 +125,54 @@ private SurveyDTO getSurveyDtoFromResponseEntity(ResponseEntity<String> response
...
-
-      return getSurveyDtoFromResponseEntity(responseEntity);
-    } catch (RestClientException e) {
-      return null;
+      survey = getSurveyDtoFromResponseEntity(responseEntity);
+    } catch (HttpClientErrorException e) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
+        return null;
+      }
+      log.error("Client error: ", e);
```
NickGrantONS: You need to rethrow the exception after logging it

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
```diff
@@ -123,46 +125,54 @@ private SurveyDTO getSurveyDtoFromResponseEntity(ResponseEntity<String> response
...
 
-      return getSurveyDtoFromResponseEntity(responseEntity);
-    } catch (RestClientException e) {
-      return null;
+      survey =  getSurveyDtoFromResponseEntity(responseEntity);
+    } catch (HttpClientErrorException e) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
+        return null;
+      }
+      log.error("Client error: ", e);
```
NickGrantONS: Again, rethrow the exception after logging it

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
```diff
@@ -123,46 +125,56 @@ private SurveyDTO getSurveyDtoFromResponseEntity(ResponseEntity<String> response
...
-      return getSurveyDtoFromResponseEntity(responseEntity);
-    } catch (RestClientException e) {
-      return null;
+      survey = getSurveyDtoFromResponseEntity(responseEntity);
+    } catch (HttpClientErrorException e) {
+      if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
+        return null;
+      }
+      log.error("Client error: ", e);
+      throw new RestClientException("There has been a client error: ", e);
```
NickGrantONS: You can just write "throw e" here... you don't need to wrap the exception

damien-lloyd:
@benjefferies @NickGrantONS It appears that the unrelated changes are as a result of formatting. I usually do a maven clean install which triggers the formatting in order to run the full test suite.

NickGrantONS:
Probably best to revert those changes

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
benjefferies: log.error("Client error statuscode={}", e.getStatusCode(), e);

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SurveySvcRestClientImpl.java
benjefferies: Maybe you could add some context via the status code here log.error("Client error statuscode={}", e.getStatusCode(), e);

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseClient.java
```diff
@@ -124,7 +101,7 @@ CollectionExerciseDTO getCollectionExercise(final UUID collexId) throws CTPExcep
           .getBody();
     } catch (UnirestException e) {
       throw new CTPException(
-          CTPException.Fault.SYSTEM_ERROR, "Failed to get collection exercise", e);
+          CTPException.Fault.SYSTEM_ERROR, "Failed to get collection exercise: %s", e);
```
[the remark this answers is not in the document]
damien-lloyd: This is so that the UnirestException is captured in the error message. Without the %s we're not capturing the actual exception being thrown.

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseClient.java
```diff
@@ -144,7 +121,7 @@ CollectionExerciseDTO getCollectionExercise(final String uriStr) throws CTPExcep
...
```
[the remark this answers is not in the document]
damien-lloyd: @benjefferies Please see above comment

src/test/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseClient.java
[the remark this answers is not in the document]
benjefferies: I don't think it give us anymore value because UnirestException doesn't implement toString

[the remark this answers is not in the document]
damien-lloyd: @benjefferies In all fairness, I added the %s to diagnose an integration test failure in collection exercise. Because it's intermittent I need more information on how to stabilise it.

PR #92, opened 2018-07-25
Fix for Collection Exercise getting stuck in EXECUTION_STARTED state

NickGrantONS:
# Motivation and Context
We have been experiencing an intermittent fault in production where Sample Units aren't getting properly verified and the Collection Exercise gets permanently stuck in EXECUTION_STARTED state.

It was theorised that this could be a thread race condition issue. During investigation, it seemed likely that a native SQL query was bypassing Spring Data JPA and Hibernate, which could be causing an incorrect count of rows to be returned versus what's in Hibernate's session cache.

This is a speculative fix, because the bug is hard to reproduce locally - it's only ever been seen in production.

# What has changed
SampleUnitRepository has a new method called countBySampleUnitGroupCollectionExercise, which uses JPA and Hibernate correctly. The native query has been removed.

SampleServiceImpl.acceptSampleUnit has been changed to use the new repo method.

A unit test class with two tests has been created for the SampleServiceImpl class, which tests the behaviour when the count is not equal to the number of expected sample units, and the behaviour when the count does equal and the state of the collection exercise should be changed.

# How to test?
Unit tests and acceptance tests have been run locally and are all passing without failures.

Additionally, there have been 3 successful runs of a sample file containing 5,000 sample units, run locally, making sure the collection exercise is fully validating without issues.

# Links
Trello: https://trello.com/c/278uXMkJ/95-rsi-1805-sample-load-and-exercise-execution-sampleserviceimpl-acceptsampleunit-logic

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImplTest.java
+++ b/src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImplTest.java
@@ -1,0 +1,113 @@
+package uk.gov.ons.ctp.response.collection.exercise.service.impl;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import java.util.UUID;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import uk.gov.ons.ctp.common.error.CTPException;
+import uk.gov.ons.ctp.common.state.StateTransitionManager;
+import uk.gov.ons.ctp.response.collection.exercise.domain.CollectionExercise;
+import uk.gov.ons.ctp.response.collection.exercise.domain.ExerciseSampleUnit;
+import uk.gov.ons.ctp.response.collection.exercise.domain.ExerciseSampleUnitGroup;
+import uk.gov.ons.ctp.response.collection.exercise.repository.CollectionExerciseRepository;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private void acceptSampleUnitWithCollex(CollectionExercise collex) throws CTPException {
+    SampleUnit sampleUnit =
+        SampleUnit.builder()
+            .withId(SAMPLE_ID.toString())
+            .withFormType("X")
+            .withSampleUnitRef("REF123")
+            .withSampleUnitType("B")
+            .withCollectionExerciseId(COLLEX_ID.toString())
```
madeye-matt: This strikes me as odd as a CollectionExercise gets passed in as a parameter - I'd expect collex.getId().toString().  I can see it works but it doesn't look right.

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImplTest.java
```diff
@@ -0,0 +1,113 @@
...
+            .build();
+
+    ExerciseSampleUnitGroup sampleUnitGroup = new ExerciseSampleUnitGroup();
+
+    when(collectRepo.findOneById(any())).thenReturn(collex);
+    when(sampleUnitGroupRepo.saveAndFlush(any())).thenReturn(sampleUnitGroup);
```
madeye-matt: I tend to use: when(sampleUnitGroupRepo.saveAndFlush(any())).then(returnsFirstArg()); for these unless you are testing an error condition

NickGrantONS: [facepalm] yes that way would be much better

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImplTest.java
NickGrantONS: have changed it

PR #93, opened 2018-07-27
Add note about migration naming

PurpleBooth:
Currently the migration names are arbitrary, this adds a structure.


PurpleBooth:
Yes, this has been discussed and agreed already. This isn't a versioniong system, it just looks like one, infact it's just a way of grouping the migrations.

PR #95, opened 2018-07-30
Remove bi/create actionplan selectors at ce creation

andrewmil:
# Motivation and Context
When we remove BI cases we will stop using `casetypeoverrides` and instead use the action plan selectors which have been recently added to the action service.
When we initially create action plans we will want to also create the appropriate action plan selectors

# What has changed
Added selectors fields to requests for creating override action plans
The selectors we are concerned with are `surveyRef`, `activeEnrolments` and `exerciseRef`

# How to test?
After running `make setup` in the acceptance tests the created action plans in the database should have it's `selectors` fields filled in

To test for a single case exercise the create collection exercise endpoint either directly or through the UI and check that action plans are created with the appropriate selectors

# Links
[Trello](https://trello.com/c/VmIr3Qwp/280-add-action-plan-selectors-when-creating-collection-exercises-m)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
@@ -1,5 +1,6 @@
 package uk.gov.ons.ctp.response.collection.exercise.client;
 
+import java.util.HashMap;
 import org.springframework.web.client.RestClientException;
 import uk.gov.ons.ctp.response.action.representation.ActionPlanDTO;
 
@@ -14,5 +15,6 @@ public interface ActionSvcClient {
    * @return ActionPlanDTO representation of the created action plan
    * @throws RestClientException for failed connection to action service
    */
-  ActionPlanDTO createActionPlan(String name, String description) throws RestClientException;
+  ActionPlanDTO createActionPlan(String name, String description, HashMap<String, String> selectors)
+      throws RestClientException;
 }
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
@@ -12,6 +12,7 @@ public interface ActionSvcClient {
    *
    * @param name name of action plan
    * @param description description of action plan
+   * @param selectors Map of selectors for actionplans as key value pairs
    * @return ActionPlanDTO representation of the created action plan
    * @throws RestClientException for failed connection to action service
    */
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
@@ -452,8 +452,8 @@ public class CollectionExerciseServiceImpl implements CollectionExerciseService 
     HashMap<String, String> selectors = new HashMap<>();
     selectors.put("exerciseRef", exerciseRef);
     selectors.put("surveyRef", survey.getSurveyRef());
-    if (!sampleUnitType.equals("H") && !sampleUnitType.equals("HI")) {
-      String activeEnrolment = Boolean.toString(sampleUnitType.equals("BI"));
+    if (!"H".equals(sampleUnitType) && !"HI".equals(sampleUnitType)) {
+      String activeEnrolment = Boolean.toString("BI".equals(sampleUnitType));
       selectors.put("activeEnrolment", activeEnrolment);
     }
     String shortName = survey.getShortName();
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
```diff
@@ -14,5 +15,6 @@
    * @return ActionPlanDTO representation of the created action plan
    * @throws RestClientException for failed connection to action service
    */
-  ActionPlanDTO createActionPlan(String name, String description) throws RestClientException;
+  ActionPlanDTO createActionPlan(String name, String description, HashMap<String, String> selectors)
```
benjefferies: It's nicer for the API to define a Map as a parameter rather than the exact type of map. Allows the consumer of the API to pass any type of Map

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/ActionSvcClientImplTest.java
```diff
@@ -62,6 +63,9 @@ public void testCreateActionPlan() {
     actionPlanDTO.setName(ACTION_PLAN_NAME);
     actionPlanDTO.setDescription(ACTION_PLAN_DESCRIPTION);
     actionPlanDTO.setCreatedBy("SYSTEM");
+    HashMap<String, String> selectors = new HashMap<>();
```
benjefferies: A nice util here would be `Collections.singletonMap("testSelector", "testValue")`

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
andrewmil: :+1: I'll add another card for that as it'd affect a few other repos

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/ActionSvcClientImplTest.java
andrewmil: Not sure if this works for `HashMap`, another reason to implement your suggestion above

MIslam92: I think it can be done like this`Collections.synchronizedMap(new HashMap("testSelector", "testValue"))` feel free to correct me if I'm wrong?

andrewmil: Had a try and couldn't get it working @MIslam92 
Might just be the way I implemented it but It'll do for now as is anyway :+1:

NickGrantONS: you're unlikely to need a synchronised collection in a unit test

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
NickGrantONS: you should be able to use the base type Map without affecting any other repos - the JSON serialisation will be identical

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -449,10 +449,17 @@ private void createOverrideActionPlan(
 
     // Create action plan with appropriate name and description
     String exerciseRef = collectionExercise.getExerciseRef();
+    HashMap<String, String> selectors = new HashMap<>();
```
NickGrantONS: in Java, always prefer to use the interface rather than the concrete type

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/ActionSvcClientImplTest.java
NickGrantONS: What Ben said is good... saves a line of code. Again, you should prefer to 'hide' the implementation by using Map as the local variable instead of the concrete type... it's Java convention/best-practice.

NickGrantONS: @MIslam92 no need to use a synchronised collection anywhere that's not multithreaded

atrozzo: I would say no need to use a an Atomic object if you are ina Thread safe context. For example a Servlet has share context so you are not offcialy in a multithreading condition ( actually you are but it s transparent ) and is not thread safe.

atrozzo: @benjefferies  not sure he needs to use a Collections.singletonMap. Is it going to be mutable or stay like this ? The main difference is not just about having the code shorted but the fact that this will create an immutable Map.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
atrozzo: Please update the method comments (Java docs ) or remove them ( i prefer we do not use code comments , the code should be self explanatory ).

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
andrewmil: A separate card will be created to change the type that action service receives from a `HashMap` to a `Map`

PR #97, opened 2018-08-02
Refactored native Postgres SQL queries to use proper JPA instead

NickGrantONS:
# Motivation and Context
Using native SQL queries hard-coded into the repositories is bad practice. JPA and Hibernate abstract our services from the underlying database vendor, and Hibernate performance - such as caching - can't be used if a native query has to be executed in the database. Also, there can be unexpected issues due to Hibernate's session management.

The improvements team is refactoring all the code to get rid of ALL the native SQL queries from the entire SDC Java codebase. This is one of several PRs.

# What has changed
A really nasty native query and an unused query have been removed from CollectionExerciseRepository. The SampleUnitDistributor.distributeSampleUnits function now uses the two Case Type repos to find the action plan, instead of the nasty native query.

Two queries have been refactored in SampleUnitRepository, so that there is no native or JPQL query.

Unit tests have been written.

# How to test?
Full acceptance test suite has been run - passing.

Some of the scenarios are quite hard to engineer - spoofing duplicate rows in the database is perhaps the only way to thoroughly test the unlikely situation where there's a duplicate Collex/Sample Unit Ref/Sample Unit type already in the DB.

This PR and the other PRs related to the refactoring should be tested with a full manual regression test to mitigate any risk of regression.

# Links
Trello: https://trello.com/c/bdqQthEj/291-replace-native-postgres-queries-with-jpa-spring-data-queries-in-java-services-l

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/distribution/SampleUnitDistributor.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/distribution/SampleUnitDistributor.java
@@ -346,7 +346,7 @@ public class SampleUnitDistributor {
     return exercise;
   }
 
-  private String getActiveActionPlanId(
+  protected String getActiveActionPlanId(
       Integer exercisefk, String sampleunittypefk, UUID surveyuuid) {
     String actionPlanId = null;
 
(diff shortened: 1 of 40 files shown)


damien-lloyd:
So long as the tests still pass, this should be OK.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/distribution/SampleUnitDistributor.java
```diff
@@ -336,4 +345,25 @@ private CollectionExercise collectionExerciseTransitionState(CollectionExercise
     }
     return exercise;
   }
+
+  protected String getActiveActionPlanId(
```
benjefferies: default I guess if it's just for package level access?

PR #80, opened 2018-07-10
Get sample summary state

benjefferies:
# Motivation and Context
A race condition could occur if a sample link record wasn't created
before the sample file hadn't finished uploading. This would prevent
collection exercises progressing to ready for review.

Instead of storing the sample unit state stored in the collection
exercise service we query sample service to get this information.

# What has changed
* Linking sample summaries can now potentially move collection exercise to READY_TO_REVIEW
* When a sample has finished loading we will query sample service to check if the sample summary is active

# How to test?
Run unit tests, integration tests and acceptance test with rm-collection-exercise-service on https://github.com/ONSdigital/rm-collection-exercise-service/pull/80 

# Links
https://trello.com/c/RcjEiELy/181-fix-linking-a-social-sample-to-a-collection-exercise
https://github.com/ONSdigital/rm-collectionexercisesvc-api/pull/17

PR #98, opened 2018-08-03
Fix collection exercise transition message

benjefferies:
# Motivation and Context
Only publish a message when the collection exercise has been
transitioned not whenever the collection exercise is updated.

This was introduced https://github.com/ONSdigital/rm-collection-exercise-service/pull/80/files

# What has changed

# How to test?
1. Run `mvn install -Ddocker.keepRunning`
2. Go to http://localhost:46672/#/queues/%2F/Collex.Transition
3. There should only be 7 events and no duplicate states for the same collection exercise id

# Links
https://github.com/ONSdigital/rm-collection-exercise-service/pull/80/files


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -660,6 +659,8 @@ public void transitionCollectionExercise(
     if (oldState != newState) {
```
PurpleBooth: Why not flip this if over so it's a guard clause?

PR #100, opened 2018-08-08
Fix social survey event dates

PurpleBooth:
Motivation and Context
======================

Currently when you create a social survey's collection exercise event
dates there is an exception. This is caused by assuming all
surveys are business surveys when creating action rules.

What has changed
================

This adds a check for if the survey is a business survey, removing the
automatic creation of action rules for non-business survey types.

How to test?
============

* Run Maven Test
* Create an event on a social survey


tej99:
[TBL123: Link action rules to action plans at create time (Create)](https://trello.com/c/Eknzy8H2/119-tbl123-link-action-rules-to-action-plans-at-create-time-create)

PR #103, opened 2018-08-10
Add client to get action plans by their selectors

andrewmil:
# Motivation and Context
A few people required this client function for what they are currently working on so this is being added independently so they have access to it

# What has changed
Added a client function for retrieving action plans from the action service using their selectors field

# How to test?
Tests should still pass
You could also exercise the function as part of a related piece of work

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
@@ -4,6 +4,7 @@ import java.time.OffsetDateTime;
 import java.util.HashMap;
 import java.util.List;
 import java.util.UUID;
+import org.springframework.web.client.HttpClientErrorException;
 import org.springframework.web.client.RestClientException;
 import uk.gov.ons.ctp.response.action.representation.ActionPlanDTO;
 import uk.gov.ons.ctp.response.action.representation.ActionRuleDTO;
@@ -23,8 +24,16 @@ public interface ActionSvcClient {
   ActionPlanDTO createActionPlan(String name, String description, HashMap<String, String> selectors)
       throws RestClientException;
 
+  /**
+   * Request list of action plans with given selectors
+   *
+   * @param collectionExerciseId collectionExerciseId to find action plans for
+   * @param activeEnrolment boolean for if sample unit has an active enrolment associated with it
+   * @return List<ActionPlanDTO> representation of the created action plan
+   * @throws HttpClientErrorException for failure to retrieve action plans
+   */
   List<ActionPlanDTO> getActionPlansBySelectors(
-      String collectionExerciseId, Boolean activeEnrolment) throws RestClientException;
+      String collectionExerciseId, Boolean activeEnrolment) throws HttpClientErrorException;
 
   /**
    * Request action rule is created.
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
@@ -24,7 +24,7 @@ public interface ActionSvcClient {
       throws RestClientException;
 
   List<ActionPlanDTO> getActionPlansBySelectors(
-      String surveyRef, String exerciseRef, Boolean activeEnrolment) throws RestClientException;
+      String collectionExerciseId, Boolean activeEnrolment) throws RestClientException;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -75,6 +81,39 @@ public ActionPlanDTO createActionPlan(
...
+    queryParams.add("activeEnrolment", activeEnrolment.toString());
+    UriComponents uriComponents =
+        restUtility.createUriComponents(appConfig.getActionSvc().getActionPlansPath(), queryParams);
+
+    ResponseEntity<List<ActionPlanDTO>> responseEntity =
+        restTemplate.exchange(
+            uriComponents.toString(),
+            HttpMethod.GET,
+            null,
+            new ParameterizedTypeReference<List<ActionPlanDTO>>() {});
```
NickGrantONS: this could throw a HttpClientException if actionSvc returns 404, in which case that runtime exception would get thrown to whatever is using this function. Would we want that behaviour, or would we want to return null?

src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImplTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.web.client.RestTemplate;
+import uk.gov.ons.ctp.common.rest.RestUtility;
+import uk.gov.ons.ctp.common.rest.RestUtilityConfig;
+import uk.gov.ons.ctp.response.action.representation.ActionPlanDTO;
+import uk.gov.ons.ctp.response.collection.exercise.config.ActionSvc;
+import uk.gov.ons.ctp.response.collection.exercise.config.AppConfig;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ActionSvcRestClientImplTest {
+
```
NickGrantONS: Maybe add a test for the 404 HttpClientException scenario too

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/ActionSvcClient.java
```diff
@@ -23,8 +24,16 @@
...
+   * Request list of action plans with given selectors
+   *
+   * @param collectionExerciseId collectionExerciseId to find action plans for
+   * @param activeEnrolment boolean for if sample unit has an active enrolment associated with it
+   * @return List<ActionPlanDTO> representation of the created action plan
+   * @throws HttpClientErrorException for failure to retrieve action plans
+   */
   List<ActionPlanDTO> getActionPlansBySelectors(
-      String collectionExerciseId, Boolean activeEnrolment) throws RestClientException;
+      String collectionExerciseId, Boolean activeEnrolment) throws HttpClientErrorException;
```
NickGrantONS: don't explicitly declare a runtime exception as if it's a checked exception

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -90,7 +90,7 @@ public ActionPlanDTO createActionPlan(
   @Override
   public List<ActionPlanDTO> getActionPlansBySelectors(
       final String collectionExerciseId, final Boolean activeEnrolment)
-    throws HttpClientErrorException {
+      throws HttpClientErrorException {
```
NickGrantONS: don't use throws for RT exceptions

src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImplTest.java
```diff
@@ -0,0 +1,134 @@
+package uk.gov.ons.ctp.response.collection.exercise.client.impl;
```
tej99: There is already an ActionSvcClientImplTest.java file with other tests for that class (the name probably should be ActionSvcRestClientImplTest.java though)

andrewmil: I think the older tests should probably be in here rather than in `test/.../service/impl` but that can probably be done as part of a wider cleanup of the tests/improvements to structure

PR #105, opened 2018-08-10
Deep if to guard clauses in acceptSampleUnit

PurpleBooth:
Motivation and Context
======================

SampleServiceImpl is one of the worst classes in codebeat.

What has changed
================

This will change the deep if to guard clauses, and add tests for the untested
routes through this function.

How to test?
============

* mvn clean install

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImpl.java
@@ -1,11 +1,5 @@
 package uk.gov.ons.ctp.response.collection.exercise.service.impl;
 
-import java.sql.Timestamp;
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/SampleServiceImpl.java
```diff
@@ -134,61 +135,61 @@ public SampleUnitsRequestDTO requestSampleUnits(final UUID id) throws CTPExcepti
...
-    } else {
+    if (collectionExercise == null) {
       log.error(
           "No CollectionExercise {} for SampleUnit Ref: {} Type: {}, FormType: {}",
           sampleUnit.getCollectionExerciseId(),
           sampleUnit.getSampleUnitRef(),
           sampleUnit.getSampleUnitType(),
           sampleUnit.getFormType());
+
+      return null;
```
benjefferies: Maybe just change the signature to be void. Looks like it's never used?

PR #107, opened 2018-08-14
Use selectors rather than case type override

tej99:
# Motivation and Context

Currently we use the case type override, these are being deprecated
soon. 

# What has changed

Remove use of case type overrides for action rules modification, and use the new action plan
selectors.
Refactor out abstractActionRuleUpdater
# How to test?

Use UI to update mps or go_live events datetime and see the matching action rule trigger datetime get updated as well in the actionrule table.

# Links

https://trello.com/c/r79j0Xlg/125-tbl123-link-action-rules-to-action-plans-at-create-time-update


src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -97,35 +109,62 @@ public ActionPlanDTO createActionPlan(
...
-      if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
-        return null;
-      }
-      throw e;
-    }
+    final MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
+    queryParams.add(SELECTOR_COLLECTION_EXERCISE_ID, collectionExerciseId);
+    queryParams.add(SELECTOR_ACTIVE_ENROLMENT, activeEnrolment.toString());
+    final ResponseEntity<List<ActionPlanDTO>> responseEntity =
+        getActionPlansByQueryParameters(queryParams);
```
NickGrantONS: you've removed the handling of 404s, so a runtime exception will be thrown in that eventuality. Is that the desired behaviour?

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -97,35 +109,62 @@ public ActionPlanDTO createActionPlan(
...
     log.debug(
         "Successfully retrieved action plan for selectors, "
             + "collectionExerciseId: {}, activeEnrolment: {}",
         collectionExerciseId,
         activeEnrolment);
+
     return responseEntity.getBody();
   }
 
+  private ResponseEntity<List<ActionPlanDTO>> getActionPlansByQueryParameters(
```
NickGrantONS: this is inconsistent with the other getters which return a list. Consider removing the ResponseEntity 'wrapper' and just returning the list inside

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -97,35 +109,62 @@ public ActionPlanDTO createActionPlan(
...
+            HttpMethod.GET,
+            null,
+            new ParameterizedTypeReference<List<ActionPlanDTO>>() {});
+
+    return responseEntity;
+  }
+
+  @Override
+  public ActionPlanDTO getActionPlanBySelectors(
+      String collectionExerciseId, Boolean activeEnrolment) throws CTPException {
```
NickGrantONS: any reason for using the Boolean class instead of the boolean primitive?

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -200,11 +240,30 @@ public ActionRuleDTO updateActionRule(
               new ParameterizedTypeReference<List<ActionRuleDTO>>() {});
     } catch (HttpClientErrorException e) {
       if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
-        return null;
+        throw new CTPException(
```
NickGrantONS: throwing this exception could be done by the caller, which would make the RestClient implementation more consistent. I think we should use the "return null on 404" pattern everywhere.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -200,11 +240,30 @@ public ActionRuleDTO updateActionRule(
...
+      throws CTPException {
+    final MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
+    queryParams.add(SELECTOR_COLLECTION_EXERCISE_ID, collectionExerciseId);
+    final ResponseEntity<List<ActionPlanDTO>> responseEntity =
+        getActionPlansByQueryParameters(queryParams);
+
+    List<ActionPlanDTO> actionPlans = responseEntity.getBody();
+
+    if (actionPlans == null) {
+      throw new CTPException(
```
NickGrantONS: again... question mark over whether we should leave the behaviour of deciding what to do on 404 to the caller for consistency

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/actionrule/ReminderActionRuleCreator.java
```diff
@@ -44,6 +49,11 @@ public void execute(
     final OffsetDateTime offsetDateTime = OffsetDateTime.ofInstant(instant, ZoneId.systemDefault());
     final CollectionExercise collectionExercise = collectionExerciseEvent.getCollectionExercise();
 
+    final ActionPlanDTO activeActionPlan =
+        actionSvcClient.getActionPlanBySelectors(collectionExercise.getId().toString(), true);
+    final ActionPlanDTO inactiveActionPlan =
```
NickGrantONS: could remove .toString in a few places if you pass by UUID instead of String. Being really pick though soz

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/EventServiceImplTest.java
```diff
@@ -1,24 +1,13 @@
 package uk.gov.ons.ctp.response.collection.exercise.service.impl;
 
 import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertThat;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
NickGrantONS: will fail style check

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/EventServiceImplTest.java
```diff
@@ -1,24 +1,13 @@
...
 import static org.mockito.Matchers.any;
-import static org.mockito.Mockito.atLeastOnce;
-import static org.mockito.Mockito.eq;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
NickGrantONS: will fail style check

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/EventServiceImplTest.java
```diff
@@ -1,24 +1,13 @@
...
 
 import java.sql.Timestamp;
 import java.time.Instant;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.List;
-import java.util.UUID;
+import java.util.*;
```
NickGrantONS: will fail style check

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
tej99: Yes after looking at action service I found this API never returns 404. Returns 204 when no actionPlans are found  ...

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
tej99: No longer needed, inlined back into getActionPlansBySelectors

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
tej99: getActionPlansBySelectors() uses a Boolean class which is where this is passed too. I can change both to use primitives? Why are primitives preferred over the class?

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
tej99: I would agree... but in this case action service gives back 204 if there are no actionRules found and gives back a 404 if the actionPlan you are trying to get the actionRules for was not found. And since this method is called getActionRulesForActionPlan and you provide the actionPlanId you're expecting there to be an actionPlan every time this method is called.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
tej99: woops this function is not getting used anymore, but the reason was the same as above ^^

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/actionrule/ReminderActionRuleCreator.java
tej99: nw cleaner 🎨

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/EventServiceImplTest.java
tej99: IDE trying to be smart

tej99:
@NickGrantONS Oh true! Changed from Boolean class to primitive time on both functions 👍

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -91,41 +103,61 @@ public ActionPlanDTO createActionPlan(
...
 
+  @Override
+  public ActionPlanDTO getActionPlanBySelectors(
+      String collectionExerciseId, boolean activeEnrolment) throws CTPException {
+
+    final List<ActionPlanDTO> actionPlans =
+        getActionPlansBySelectors(collectionExerciseId, activeEnrolment);
+
+    if (actionPlans == null) {
+      throw new CTPException(
```
NickGrantONS: How is this ever going to be null? The behaviour of getActionPlansBySelectors has been changed to throw a runtime exception in the event of a non 2XX HTTP response code

tej99: 205 No content is returned when no action plans exist so responseEntity.getBody() would be null.

tej99:
[TBL123: Link action rules to action plans at create time (Update)](https://trello.com/c/r79j0Xlg/125-tbl123-link-action-rules-to-action-plans-at-create-time-update)

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -91,41 +103,61 @@ public ActionPlanDTO createActionPlan(
...
   @Override
   public List<ActionPlanDTO> getActionPlansBySelectors(
-      final String collectionExerciseId, final Boolean activeEnrolment) {
+      final String collectionExerciseId, boolean activeEnrolment) {
     log.debug(
         "Retrieving action plan for selectors, " + "collectionExerciseId: {}, activeEnrolment: {}",
         collectionExerciseId,
         activeEnrolment);
 
-    MultiValueMap<String, String> queryParams = new LinkedMultiValueMap<>();
```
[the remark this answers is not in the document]
NickGrantONS: it's what RestUtility.createUriComponents expects.

I suppose we could improve the 'helper' class.

PR #112, opened 2018-08-20
Add authentication header & remove unnecessary survey calls

tej99:
# Motivation and Context

Call to action service was being done without an authentication header
resulting in 401s on event update. This pr adds the header into the
request. Also a number of unnecessary calls are being made to survey
service, changed the logic order so the calls are not being made
unless it is needed.

# What has changed

Added authentication header to action service call
Changed order for calling survey service to after the event tag check.

# How to test?

mvn test
ATs

# Links

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
@@ -225,13 +225,14 @@ public class ActionSvcRestClientImpl implements ActionSvcClient {
             appConfig.getActionSvc().getActionRulesForActionPlanPath(), null, actionPlanId);
 
     ResponseEntity<List<ActionRuleDTO>> response;
+    HttpEntity httpEntity = restUtility.createHttpEntity(null);
 
     try {
       response =
           restTemplate.exchange(
               uriComponents.toUri(),
               HttpMethod.GET,
-              null,
+              httpEntity,
               new ParameterizedTypeReference<List<ActionRuleDTO>>() {});
     } catch (HttpClientErrorException e) {
       if (e.getStatusCode() == HttpStatus.NOT_FOUND) {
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/actionrule/GoLiveActionRuleCreator.java
```diff
@@ -28,21 +28,21 @@ public GoLiveActionRuleCreator(
...
+    if (!isGoLive(collectionExerciseEvent)) {
       return;
     }
 
-    if (!isGoLive(collectionExerciseEvent)) {
+    final CollectionExercise collectionExercise = collectionExerciseEvent.getCollectionExercise();
+
+    final SurveyDTO survey = surveyService.getSurveyForCollectionExercise(collectionExercise);
+
+    if (survey.getSurveyType() != SurveyDTO.SurveyType.Business) {
```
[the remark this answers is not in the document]
tej99: I was thinking these classes will also be used for social surveys as well just with a ternary operation when choosing the action rule type. (as you suggested before)

tej99:
@benjefferies forgot to update the unit tests to check that the HTTPEntity is being passed in, all done now 👍

src/main/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/ActionSvcRestClientImpl.java
```diff
@@ -225,13 +225,14 @@ public ActionRuleDTO updateActionRule(
             appConfig.getActionSvc().getActionRulesForActionPlanPath(), null, actionPlanId);
 
     ResponseEntity<List<ActionRuleDTO>> response;
+    HttpEntity httpEntity = restUtility.createHttpEntity(null);
```
PurpleBooth: This is the line that adds the authentication header?

benjefferies: It's not very obvious... I've create https://github.com/ONSdigital/rm-common-service/pull/14/files to help

PR #121, opened 2018-09-03
Adjust reminder event validation

tej99:
# Motivation and Context

Currently reminder events that are in the past can not be updated,
which is a valid validation, but this check should only apply if the
collection exercise has been executed. This PR adds a check for CE
state before validating the reminder event has passed. Allowing a user
to update reminder events that have passed but not actually created any
notification files/emails.

# What has changed

Only validate existing reminder event hasn't passed if collection exercise has been executed.

# How to test?

Ran unit tests, integration tests and acceptance tests.

# Links


tej99:
[Add events validation for creating and updating events](https://trello.com/c/Ig1gQMls/171-add-events-validation-for-creating-and-updating-events)

PR #108, opened 2018-08-15
Remove bi/stop creating bi cases

andrewmil:
# Motivation and Context
This PR stops `BI` cases being created at collection exercise execution
We also now choose which action plan a sample unit should be on using the action plan selectors endpoint in action service

# What has changed
Code now assumes that we'll only have one sample unit per sample unit group
Removed logic for creating `BI` cases
Stop using casetypeoverrides to find a sample unit's action plan id. Now use action plan selectors endpoint in action service

# How to test?
`mvn clean install` should pass
Executing a collection exercise with an already enrolled sample unit should not create `BI` cases

Acceptance tests will fail until other work in the case service is complete

# Links
[Trello](https://trello.com/c/ngpjOB5c/15-only-create-one-case-per-sample-unit-collection-exercise-execution-l)

!!! DO NOT MERGE !!!

PR #122, opened 2018-09-04
Remove unnecessary use of interface

benjefferies:
# Motivation and Context
YAGNI. The interface is not needed because it's unlikely there will be any
other implementations

# What has changed
Remove interface

# How to test?
Tests should still pass


benjefferies:
Already done by @NickGrantONS on https://github.com/ONSdigital/rm-collection-exercise-service/pull/108

PR #125, opened 2018-09-07
Fix race condition on sample size

NickGrantONS:
# Motivation and Context
Presently the Collection Exercise service doesn't know how many sample units it expects to be sent by the Sample service until it requests the sample units to be sent, which causes a race condition - if the sample units start arriving before Collex has managed to store the expected number to the database.

This was causing intermittent failures in the CI pipeline and in production, and a temporary fix was to ignore the count check if we didn't have an expected value, but this fix was less than ideal, because the race condition still potentially existed for very small sample sizes.

This fix is intended to close all the loopholes by asking for the number of expected sample units _before_ they start to be sent by the Sample service. This is **one of three** related PRs.

# What has changed
Before instructing Sample service to send the sample units, while handling a `CollectionExerciseExecutionEndpoint.requestSampleUnits` request, we first request the total number of sample units we expect to receive from the Sample service, using a new REST API call.

We have to ensure that the transaction is committed before any Rabbit messages start arriving, so the update to the database for the count is done in its own transaction.

# How to test?
This is an intermittent bug which is hard to reproduce. Try uploading some samples to make sure there's no regression

# Links
Trello: https://trello.com/c/DaKcP1zA/379-race-condition-fix-for-sample-collex-exchange-of-sample-data

### DIFF
--- a/src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SampleSvcRestClientImplTest.java
+++ b/src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SampleSvcRestClientImplTest.java
@@ -99,7 +99,7 @@ public class SampleSvcRestClientImplTest {
     given(
             restTemplate.exchange(
                 any(URI.class),
-                eq(HttpMethod.POST),
+                eq(HttpMethod.GET),
                 any(HttpEntity.class),
                 eq(SampleUnitsRequestDTO.class)))
         .willReturn(responseEntity);
--- a/src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SampleSvcRestClientImplTest.java
+++ b/src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SampleSvcRestClientImplTest.java
@@ -6,6 +6,7 @@ import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 
 import java.net.URI;
+import java.util.Collections;
 import java.util.UUID;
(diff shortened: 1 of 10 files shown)


src/test/java/uk/gov/ons/ctp/response/collection/exercise/client/impl/SampleSvcRestClientImplTest.java
```diff
@@ -78,4 +84,31 @@ public void getSampleSummariesBadHttpCode() {
...
+
+    // Given
+    given(appConfig.getSampleSvc()).willReturn(sampleSvc);
+    given(sampleSvc.getRequestSampleUnitCountPath()).willReturn("test/path");
+    given(responseEntity.getStatusCode()).willReturn(HttpStatus.OK);
+    given(responseEntity.getBody()).willReturn(response);
+    given(
+            restTemplate.exchange(
+                any(URI.class),
+                eq(HttpMethod.POST),
```
tej99: GET?

tej99: I'm surprised this passed in travis :/

tej99:
Looks like travis isn't configured properly, build passed even though we have a failing test 
```[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   SampleSvcRestClientImplTest.getSampleUnitSizeHappyPath:109 » NullPointer
[INFO]
[ERROR] Tests run: 287, Failures: 0, Errors: 1, Skipped: 0```

PR #126, opened 2018-09-11
Add optional parameter to REST API to allow filter by only live collexes

NickGrantONS:
# Motivation and Context
We are currently filtering in the front end, to display only the live collection exercises. By adding an optional parameter to the REST API, we can simplify the front-end code and reduce the amount of data being transferred from DB -> Collex Service -> Front-end.

# What has changed
Added optional param to `CollectionExerciseEndpoint.getCollectionExercisesForSurvey`.

# How to test?
There's a new unit/REST test which covers this change.

To manually test, send a GET to `http://localhost:8145/collectionexercises/survey/{id}?liveOnly=true` and you should only receive collection exercises which are in the LIVE state.

# Links
Trello: https://trello.com/c/n7EcEA9y/344-add-ability-to-collection-exercise-to-only-return-live-collection-exercises-for-a-survey-m

### DIFF
--- a/API.md
+++ b/API.md
@@ -18,6 +18,7 @@ This page documents the Collection Exercise service API endpoints. Apart from th
 
 ## Get Collection Exercises for Survey
 * `GET /collectionexercises/survey/{survey_id}` will return a list of known collection exercises for the survey id.
+* `GET /collectionexercises/survey/{survey_id}?liveOnly=true` will return a list of LIVE collection exercises for the survey id.
 
 ### Example JSON Response
 ```json
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
@@ -35,6 +35,7 @@ import uk.gov.ons.ctp.response.collection.exercise.repository.CaseTypeOverrideRe
 import uk.gov.ons.ctp.response.collection.exercise.repository.CollectionExerciseRepository;
 import uk.gov.ons.ctp.response.collection.exercise.repository.SampleLinkRepository;
 import uk.gov.ons.ctp.response.collection.exercise.representation.CollectionExerciseDTO;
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/impl/CollectionExerciseServiceImpl.java
```diff
@@ -102,6 +103,12 @@ public CollectionExerciseServiceImpl(
     return this.collectRepo.findBySurveyId(UUID.fromString(survey.getId()));
   }
 
+  @Override
+  public List<CollectionExercise> findCollectionExercisesBySurveyIdAndState(
+      UUID surveyId, CollectionExerciseState state) {
+    return collectRepo.findBySurveyIdAndState(surveyId, state);
```
MIslam92: Codecov is complaining this is not covered in a test might also be worth adding one for it

NickGrantONS: a unit test for a 1-liner seems excessive

NickGrantONS: we're going to refactor away the stupid wrapper functions around repo methods

MIslam92: I thought bout that too but just codecov being codecov but yeah if we're removing them wrapper functions thats cool 👍

API.md
```diff
@@ -18,6 +18,7 @@ This page documents the Collection Exercise service API endpoints. Apart from th
 
 ## Get Collection Exercises for Survey
 * `GET /collectionexercises/survey/{survey_id}` will return a list of known collection exercises for the survey id.
+* `GET /collectionexercises/survey/{survey_id}?liveOnly=true` will return a list of LIVE collection exercises for the survey id.
```
tej99: 👍

PR #137, opened 2018-10-24
Initial changes reflecting party dropping sample_unit_type from endpo…

AndrewTorrance:
# Motivation and Context
Party endpoints changed to not accept sample unit type 

# What has changed
Party end point definition 

# How to test?
Acceptance tests

# Links
 https://trello.com/c/I5lc4gDq


AndrewTorrance:
The is PR has become stale and relies on BI cases being removed which may lead to further work. Hence closing this with a view to reappraising AFTER BI cases removed.

PR #141, opened 2019-03-07
587 caching performance improvements

ryangrundy7:
# Motivation and Context
To help improve the performance of the collection exercise validation and distribution, I've added cacheable annotations to the collection instrument and action plans endpoints to reduce the amount of calls it makes. From testing locally with a 5000 sample i've seen about a 1-2 minute improvement to validating the collection exercise and a slight improvement of the distribution to the case service. 

The performance could be even more improved by decreasing the `validation-schedule-delay-milli-seconds` and `distribution-schedule-delay-milli-seconds` time from 1 minute to something lower. Testing the effects this would have on the system locally is difficult, so it'd be best to test this in the performance environment when it's ready to go.

# What has changed 
- Added Cacheable annotations to `actionplans` and `collectioninstrument`

# How to test?
- In docker the `validation-schedule-delay-milli-seconds` and `distribution-schedule-delay-milli-seconds` are set to 1 second so to make it like production, you should change this to `60000`
- Run the tests
- Add some log lines inside `getActionPlanBySelectorsBusiness` and `requestCollectionInstruments` and execute a collection exercise, it should use a cached version and they evict the cache after about a minute. 

# Links
[Trello card](https://trello.com/c/lA9ni5qG)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/CollectionExerciseApplication.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/CollectionExerciseApplication.java
@@ -239,7 +239,7 @@ public class CollectionExerciseApplication {
   public void cacheEvict() {
     /* This is getting rid of the cached entries in case anything's been changed. We imagine
     that
-        * the maximum of 1 minute delay to seeing changes reflected in the case service will not
+        * the maximum of a 1 minute delay to seeing changes reflected in the collection exercise service will not
         * cause any issues*/
   }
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/CollectionExerciseApplication.java
```diff
@@ -44,6 +51,9 @@
 @ComponentScan(basePackages = {"uk.gov.ons.ctp.response"})
 @EnableJpaRepositories(basePackages = {"uk.gov.ons.ctp.response"})
 @EntityScan("uk.gov.ons.ctp.response")
+@EnableScheduling
+@EnableCaching
+@EnableAsync
```
NickGrantONS: Why enable async? We're not using it anywhere so far as I can see, and it can have quite unpredictable results sometimes.

ryangrundy7: Will remove it 👍

PR #142, opened 2019-03-12
Refactor sample validation

NickGrantONS:
# Motivation and Context
The 'validation' of the samples was spaghetti code and highly inefficient.

Given the need to load large samples for ASHE etc. it seemed prudent to refactor all the code that the lies on the sample load pipeline, to make it more performant and reliable.

In the course of the refactoring, a number of bugs were spotted, such as inability to recover from momentary connection problems with the Survey, Collection Instrument and Party services - these should retry without manual intervention.

The code was inefficient, badly performant, hard to understand, maintain, and it did not handle errors well.

# What has changed
Major refactoring of the `ValidateSampleUnits` class.

Sadly it's still not quite honouring the single responsibility principle, but the code is a lot less complex and should make future refactoring easier.

Substantial improvement to performance and error handling.

# How to test?
Load a big sample. Make sure the sample validates OK and the collection exercise changes to "Ready for Live" status.

# Links
Trello: https://trello.com/c/eAmhl3iB/729-use-stream-and-multithreading-in-validatesampleunits-to-rapidly-validate-the-samples-13

PR #146, opened 2019-05-01
BUG: same collection instrument was being used for every sample in a collex

NickGrantONS:
# Motivation and Context
A bug had been introduced as part of performance improvement work, intended to speed up the amount of time it takes to validate a sample. The bug was putting the same collection instrument onto every sample, regardless of form type.

# What has changed
The cache has been enhanced so that it caches collection instruments down to a form type granularity.

# How to test?
Load a sample with multiple form types. Check that the resulting cases have the correct collection instruments associated against them.

# Links
Trello: https://trello.com/c/sySrEfGS
Root cause of bug: https://github.com/ONSdigital/rm-collection-exercise-service/pull/142

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnits.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnits.java
@@ -84,7 +84,11 @@ public class ValidateSampleUnits {
     // the sample service are state transitioned once they've got all their sample units.
     updateExecutingCollectionExercises();
 
-    Map<CollectionExercise, Optional<UUID>> collectionInstrumentMap = new HashMap<>();
+    // This is a hand-rolled cache which stores all the collection instruments which are needed
+    // to be added onto the samples which are being 'validated'. The map allows a collection
+    // exercise and a sample form type to be mapped to the correct collection instrument. It
+    // provides a very significant performance boost when 'validating' large samples (e.g. 100k+)
+    Map<CollectionExercise, Map<String, Optional<UUID>>> collectionInstrumentMap = new HashMap<>();
 
     try (Stream<ExerciseSampleUnit> sampleUnits =
         sampleUnitRepo.findBySampleUnitGroupCollectionExerciseStateAndSampleUnitGroupStateFK(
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnits.java
```diff
@@ -84,25 +84,36 @@ public void validateSampleUnits() {
     // the sample service are state transitioned once they've got all their sample units.
     updateExecutingCollectionExercises();
 
-    Map<CollectionExercise, Optional<UUID>> collectionInstrumentMap = new HashMap<>();
+    // This is a hand-rolled cache which stores all the collection instruments which are needed
+    // to be added onto the samples which are being 'validated'. The map allows a collection
+    // exercise and a sample form type to be mapped to the correct collection instrument. It
+    // provides a very significant performance boost when 'validating' large samples (e.g. 100k+)
+    Map<CollectionExercise, Map<String, Optional<UUID>>> collectionInstrumentMap = new HashMap<>();
```
mfcrocker: A Map of Maps feels like it's going to be a huge memory hog with large samples - have you tested it out?

src/test/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnitsTest.java
```diff
@@ -133,6 +139,100 @@ public void testValidateSampleUnits() throws CTPException {
         .transitionCollectionExercise(collectionExercise, CollectionExerciseEvent.VALIDATE);
   }
 
+  @Test
+  public void testValidateSampleUnitsMultipleCollectionInstruments() throws CTPException {
+    // Given
```
AndrewTorrance: This would benefit  from a description showing why we need this tests , and how this test satisfies that . I think such a description would be too long to to fit within a modified function name .

src/main/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnits.java
NickGrantONS: Usually there will only be one sample file being processed at any one time with a handful of form types. Because the sample is generally only for one collex, about the most this map of maps would ever hold would be about 30 UUIDs (e.g. MBS)

src/test/java/uk/gov/ons/ctp/response/collection/exercise/validation/ValidateSampleUnitsTest.java
NickGrantONS: added comment

PR #152, opened 2019-12-10
New endpoint

mfcrocker:
# Motivation and Context
This will enable future work on the dashboard

### DIFF
--- a/API.md
+++ b/API.md
@@ -629,3 +629,6 @@ json
 ```
 
 An `HTTP 404 Not Found` status code is returned if an event could not be found.
+
+## Get Events linked to list of Collection Exercises
+* `GET /collectionexercises/events?ids={a,b}` will return the events linked to the specified collection exercise IDs.
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
@@ -773,6 +773,31 @@ public class CollectionExerciseEndpoint {
     return ResponseEntity.ok(result);
   }
 
+  @RequestMapping(value = "/events/{ids}", method = RequestMethod.GET)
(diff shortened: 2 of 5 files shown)


API.md
```diff
@@ -628,4 +628,7 @@ json
 ]
 ```
 
-An `HTTP 404 Not Found` status code is returned if an event could not be found.
\ No newline at end of file
+An `HTTP 404 Not Found` status code is returned if an event could not be found.
+
+## Get Events linked to list of Collection Exercises
+* `GET /collectionexercises/events?ids={a,b}` will return the events linked to the specified collection exercise IDs.
```
insacuri: Can you include an example?  Using this, I would've put the curly braces in but the unit test says otherwise.
Can you also remove the curly braces from this example as they're not needed and are misleading

src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -773,6 +773,31 @@ private CollectionExerciseDTO getCollectionExerciseDTO(
     return ResponseEntity.ok(result);
   }
 
+  @RequestMapping(value = "/events", method = RequestMethod.GET)
+  public ResponseEntity<HashMap<UUID, List<EventDTO>>> getMultipleCollectionExerciseEvents(
+      @RequestParam("ids") final List<UUID> ids) throws CTPException {
+    List<EventDTO> result =
```
jterryons: I think this can be simplified. The for loop looks like a groupingBy operation. Would have to change the method signature to return a Map instead of a HashMap though.
```
Map<UUID, List<EventDTO>> response =
        this.eventService
            .getEvents(ids)
            .stream()
            .map(EventService::createEventDTOFromEvent)
            .collect(Collectors.groupingBy(EventDTO::getCollectionExerciseId,
                Collectors.toList()));

    return ResponseEntity.ok(response);
```

PR #212, opened 2020-11-19
created database migration for deleting tables and procedures

anwilkie:
# Motivation and Context
The MI reporting system isn't used in RASRM, so there is no need for it and any related tables/ procedures. From what I could tell, there was no Java code relating to these tables/procedures on the service itself.

# What has changed
* Created database migration to drop three tables (`report`, `reporttype` and `messagelog`) and two procedures (`generate_collectionexercise_mi` and `logmessage`).
* Removed a line from the `improvements.md` about removing the `report` and `reporttype` tables, as that will be done here. Also removed a mention of the `messagelog` table.

# Links
[Trello card](https://trello.com/c/ZE7RDTVg)

### DIFF
--- a/IMPROVEMENTS.md
+++ b/IMPROVEMENTS.md
@@ -139,8 +139,6 @@
    Holds association between collection exercise and action plan id
    #### `samplelink`
    Holds collection-exercise and sample summary id
-   #### `messagelog`
-   Holds job id and message information.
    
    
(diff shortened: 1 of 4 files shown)


jterryons (approved):

warrenbailey:
This won't work as it's missing from the changelog master

PR #218, opened 2020-11-30
Process events endpoint

insacuri:
# Motivation and Context
Needs preprod and prod config added to this service for case before it's merged.

# What has changed

# How to test?
- Test with actionDeprecated = false.  The changes in the sampleUnitDistributor where it now sends an `activeEnrolment` boolean instead of an `actionPlanId` means the full end to end journey doesn't fully work.  The test with it switched off will ensure that it still works in downstream environments until we're ready to switch it on.
- It's possible to test it a bit with actionDeprecated = true, you'll see the SCHEDULED events getting picked up and set to processed (but not acted on as that doesn't exist yet)

# Links


mfcrocker:
Please reopen this PR after rebasing to bring in the Github Actions changes

PR #223, opened 2020-12-08
Event status update existing

jdfox:
# Motivation and Context
The events that have currently been created have a NOT_SET state. When we do the switch to the new version that talks to case instead of action, we'll need to update existing events to either be:
- PROCESSED (in the case of past events)
- SCHEDULED (in the case of future events)

# What has changed
Liquibase updates to `collectionexercise.event`

# How to test?

# Links
- https://trello.com/c/pmXZmR5Y


insacuri (changes requested):
You don't want to include the file in the changelog master yet as it'll run once it gets into an environment

insacuri (approved):

PR #242, opened 2021-03-08
Remove deprecateAction feature flag and cleanup unused code

insacuri:
# Motivation and Context
A deprecateAction feature flag was added to collection exercise to facilitate the change in way that actions were handled in the system.  

This PR removes a good bit of the now unused functionality but not all of it.  We're doing this in stages to make the PRs easier to review and to reduce risk.

# What has changed

# How to test?

# Links


src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/ExerciseSampleUnit.java
```diff
@@ -70,22 +70,6 @@ public SampleUnitParent toSampleUnitParent(
     return parent;
   }
 
-  // To be removed once actionplans are deprecated
```
akssinhaamit, line 73: Not sure if I understand it right. But we will need collection exercise id,  sample unit type and sample unit ref. to be propagated to case. Is this right?

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/CollectionExerciseService.java
```diff
@@ -419,172 +413,6 @@ private CollectionExercise newCollectionExerciseFromDTO(CollectionExerciseDTO co
...
-  }
-
-  /**
-   * Create case type default for action plan, survey and sample unit type if not already exists
-   *
-   * @param survey DTO Representation of survey
-   * @param sampleUnitType Sample Unit Type
-   * @param actionPlan DTO Representation of action plan
-   */
-  private void createCaseTypeDefault(
```
akssinhaamit, line 497: Again not sure but don't we want to create case type based on active enrolment if survey and sample unit type not already exists?

akssinhaamit (commented):
Just two small queries but the PR looks ok.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/ExerciseSampleUnit.java
insacuri: There are 2 functions called `toSampleUnitParent` that have different signatures.  The one I'm deleting takes an actionPlanId and collectionExerciseId.
The one that's remaining takes an boolean activeEnrolment and a collectionExerciseId.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/CollectionExerciseService.java
insacuri: umm, can you show me how it's used in the new world?  I took another look and all the caseType stuff refers to actionPlans and is only used in relation to actionplans (which don't exist anymore)

akssinhaamit (approved):

PR #256, opened 2021-06-11
removing unused rabbit queues from collection exercise

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/svAgAyV2/733-remove-unused-rabbit-queue-rm-collection-exercise-service


src/main/resources/springintegration/flows.xml
```diff
@@ -7,11 +7,6 @@
   http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd">
 
     <int:annotation-config/>
-	<import resource="case-outbound-flow.xml"/>
-    <import resource="collex-event-messager-outbound.xml"/>
-    <import resource="collex-event-messager-inbound.xml"/>
```
warrenbailey, line 12: are we able to delete these files as well as removing them from here?

warrenbailey (changes requested):

warrenbailey:
looks like you need to run `mvn fmt:format`

src/main/resources/springintegration/flows.xml
creativeappdevamit: yes.. don't know why my last commit didn't fo in.. I did remove it but did not check in.

creativeappdevamit: done

warrenbailey (approved):

PR #257, opened 2021-06-14
Removing rabbit completely from collex

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/ITyML5G9/732-replace-rabbit-with-rest-call-ras-collection-instrument-to-rm-collection-exercise-service


src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseInstrumentReceiverEndpoint.java
```diff
@@ -0,0 +1,44 @@
...
+  @RequestMapping(value = "/link", method = RequestMethod.POST)
+  public ResponseEntity<?> collectionInstrumentLink(
+      final @RequestBody @Valid CollectionInstrumentMessageDTO collectionInstrumentMessageDTO)
+      throws CTPException {
+    log.with(collectionInstrumentMessageDTO.getInstrumentId())
+        .with(collectionInstrumentMessageDTO.getExerciseId())
+        .info("Consumed message");
+
+    UUID collectionExerciseId = collectionInstrumentMessageDTO.getExerciseId();
+
```
warrenbailey, line 39: just so i'm clear, collection instrument sends the instrument id to CE but it does nothing with it?

warrenbailey (changes requested):
A couple of questions

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/CollectionInstrumentLoadedReceiver.java
```diff
@@ -21,8 +20,8 @@
   public void acceptSampleUnit(CollectionInstrumentMessageDTO message) throws CTPException {
     log.with(message).debug("Consumed message");
 
-    UUID collexId = message.getExerciseId();
+    // UUID collexId = message.getExerciseId();
```
warrenbailey, line 24: Do we need this class anymore if all it does is log?

src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseInstrumentReceiverEndpoint.java
creativeappdevamit: Seems like it only uses it for logging against which collection instrument id its doing the transition change.
The actual instrument Id is registered as a part of the sample unit communication from sample service.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/CollectionInstrumentLoadedReceiver.java
creativeappdevamit: We don't will take it out.

warrenbailey (approved):

PR #263, opened 2021-08-18
replacing pubsub impl to spring integration

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/cZwIJ450/870-implement-spring-integration-for-pubsub-between-collection-exercise-and-case-service


insacuri (changes requested):
I'm almost certain that all of this can go once the v2 stuff is live.  From a code standpoint it looks good and could go in; Though I think we should close this PR as we're imminently going to be deleting the code that handles samples going to CE and publishing samples to case from CE

creativeappdevamit:
> I'm almost certain that all of this can go once the v2 stuff is live. From a code standpoint it looks good and could go in; Though I think we should close this PR as we're imminently going to be deleting the code that handles samples going to CE and publishing samples to case from CE

PR #266, opened 2021-08-31
Removing Sample V1 code

warrenbailey:
# What and why?
Now that sample v2 is enabled and sample is responsible for validation and distributing the sample file a lot of code can be deleted in collection exercise including 2 cron jobs and 4 database tables. This should help to simplify the service considerably as it is now only responsible for a collection exercise and all the events that happen on that.

# How to test?
Deploy to an environment, ensure the acceptance test passed and then create a new collection exercise and set ready for live

# Trello
https://trello.com/c/S5Jz26gc/441-s42-ce-sample-remove-feature-flag-code-from-collection-exercise


src/main/java/uk/gov/ons/ctp/response/collection/exercise/endpoint/CollectionExerciseEndpoint.java
```diff
@@ -970,18 +966,6 @@ private CollectionExerciseDTO getCollectionExerciseDTO(
     // is not validated and passed on verbatim
     collectionExerciseDTO.setSurveyId(collectionExercise.getSurveyId().toString());
 
-    // If we are in the failed validation state, then there should be validation error so go look
```
warrenbailey, line 973: this needs to go back in now

insacuri:
```
request-collection-instruments: /collection-instrument-api/1.0.2/collectioninstrument
```
in the application.yml and the associated code config needs to be removed now that nothing uses it

insacuri (changes requested):
One small change but this is looking good

insacuri (approved):

PR #268, opened 2021-09-06
Check cases

insacuri:
# What and why?

There was an issue that occurred mostly around automated tests but in theory could happen to a user if they got unlucky with timings or case was being exceptionally slow.  
The problem was that Sample would tell collection exercise it was done distributing messages to case and collection exercise would set the exercise live (if the go_live date was in the past) without case having processed all of the messages that sample had sent.

This PR makes it so that before it processes any event, it ensures that known sample size (stored in collection exercise) and the number of cases for that collection exercise (stored in case) match.  If they don't it'll wait until the next pass of the process events cronjob and check next time.

Additionally, there was a  'cheat' removed that shouldn't affect how things work in a real world scenario.  When sample finished distributing, it tells collection exercise that it's successful.  Collection  exercise would  then see if the go_live date was in the past; if so, then it would skip the READY_FOR_LIVE phase and go directly to live.
By removing this, we make the process of  getting an exercise live simpler as there is only 1 place that it can go live (via the go_live event, which is processed by a cronjob).  We also only have to do the sample size vs cases check in 1 place, rather then 2.

The cost to this is this acceptance tests will be a little slower, as it needs to wait for the once a minute process events cronjob.  But in a real world scenario, the MPS and go live will be hours/days/weeks in advance so there will be no change to how this works.

# How to test?

Run the acceptance tests (with the `refresh-instead-link-click` branch).  If you tail the collection exercise logs, you will see it remain ready for live until the cronjob comes and processes the `go_live` event, after which the exercise will be live.  The logs  will also show the sample size vs case check.

# Trello


warrenbailey (approved):
Looks good, a couple of suggestions for code readability but what you have will work.

warrenbailey (approved):

PR #295, opened 2022-06-22
Remove redis from collection exercise

insacuri:
# What and why?

During the java 17 upgrade, we noticed the redisson package needed updating.  After thinking about it, we might not need redis in collection exercise as that was probably used as part of the old style of sample load/case creation.  This PR will remove all mentions of it to make and then we can test that everything still works as expected.

# How to test?

- Deploy environment with this PR in it
- Either
  - Run acceptance tests (to set up an exercise and have someone sign up for it), or
  - Create a collection exercise in your environment, set it live and enrol on it.
- After this, check response ops pages to make sure nothing breaks.  Verify the response chasing report still works
- Also check frontstage as a respondent.

Nothing should be different than it normally is as everything that was deleted should've only been used by the old sample load system.  Also, now that we use pubsub, that rabbitmq config (in the 'messaging' key) should be fine to delete as well.

# Trello


pom.xml
```diff
@@ -196,6 +184,13 @@
             <version>1.2.5</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.jodd</groupId>
+            <artifactId>jodd-util</artifactId>
```
pricem14pc, line 189: Had this previously been a dependency of `redission`?

pricem14pc: Absolutely.. that's for another PR. It looks like our sole usage of it is to treat a sequence of whitespace the same as null or empty.

pricem14pc (approved):
- successfully tested deployment (albeit with existing `main` branch `ras-rm-config`)
- workload configuration `redis-config` no longer exists as expected
- container environment variables `REDIS_HOST`, `REDIS_PORT`, `REDISSON_CONFIG_ADDRESS` no longer exist as expected
- acceptance tests passed successfully
- manual round of collection exercise creation through to live, generating UACs, registering new accounts, launching EQ
- ROps activities behaving as expected (search, download progress report etc)

PR #300, opened 2022-10-27
Update publish dev chart

LJBabbage:
# What and why?
Currently on a push to a PR each repo builds to latest in onsrasrm-management, however this is the default artifact for all builds when using spinnaker. This means if any change to the charts are made and a PR is created it can and will break dev for everyone. 

# How to test?
Check the build (Publish dev Chart) and make sure it is writing naming the artifact correctly, it should be collection-exercise-pr-300.tgz. You can also check on GCP in ons-rasrmbs-management that the artifact was written out at the right time
# Trello

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,6 +155,7 @@ jobs:
       - name: Publish dev Chart
         if: github.ref != 'refs/heads/main'
         run: |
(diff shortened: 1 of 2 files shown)


pricem14pc (approved):
As per the test steps, the PR GHA is correctly naming the tgz for the pr-300 and the artifact is pushed to the GCS named correctly.

PR #304, opened 2022-11-18
BUG: Setting CE to live with wrong CI causes errors

SteveScorfield:
# What and why?
It was found that when uploading a SEFT CI and a sample, with different form types, they failed validation (as expected). However, when trying to correct the issue by uploading the correct files, an error was thrown. It was being caused by the state of the CE being set to failed validation and needs to be changed to a state of scheduled.

I have changed the code, so that the failed validation state of the CE reverts back to a scheduled event. This will allow editing of the uploads without any exceptions/errors.

# How to test?

1. Set up a CE with correct outlined times and dates
2. Uploaded SEFT form type 0001
3. Uploaded sample looking for form type 0002
4. Hit the “set ready for live” button or if didn’t appear, hit the "edit date" MPS date and press “Save” the button should now be visible 
5. I pressed “Set as ready for live” and selected “OK” on the popup confirmation window.
6. Went into “Replace sample file”  and pressed “replace sample file”
7. Sample should have loaded with no issues

# Trello
https://trello.com/c/G1zMQCAB/1964-bug-setting-ce-to-live-with-wrong-ci-causes-errors-track-time

### DIFF
--- a/_infra/helm/collection-exercise/Chart.yaml
+++ b/_infra/helm/collection-exercise/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 13.0.6
+version: 13.0.7
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 13.0.6
+appVersion: 13.0.7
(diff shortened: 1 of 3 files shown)


SteveScorfield:
[BUG: Setting CE to live with wrong CI causes errors (track time)](https://trello.com/c/G1zMQCAB/1964-bug-setting-ce-to-live-with-wrong-ci-causes-errors-track-time)

matthew-robinson-ons (approved):
Pr functions correctly and allows for correct replacement of sample files after uploading incorrect type.

PR #308, opened 2023-02-03
add redis client

JacobHarrisonONS:
# What and why?
Add functionality to invalidate redis cache when an exercise goes live

# How to test?
Create collection, when it goes live should be able to refresh page and it will appear in todo list. The default expiry is 10 minutes so you may want to deploy https://github.com/ONSdigital/ras-frontstage/pull/886 giving you 30minutes to create and go live exercise. Can test without this code using that 30minutes if you want to confirm that it doesn't invalidate cache. 

# Trello
https://trello.com/c/nKZNVODc/2440-ability-to-invalidate-collection-exercise-redis-cache-from-rops


src/main/java/uk/gov/ons/ctp/response/collection/exercise/lib/common/redis/SessionResolver.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.ons.ctp.response.collection.exercise.lib.common.redis;
+
+import com.godaddy.logging.Logger;
+import com.godaddy.logging.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+@Component
+public class SessionResolver {
```
insacuri, line 9: The naming of this feels wrong to me.  This isn't resolving a session, it's just a key deleter (in its current state).   How is this class going to be used in the future?  Do you have one of these classes per thing you want to do? (i.e., one for deleting, another for adding, etc) or per group of thing? (i.e., one for setting/removing keys from exercise endpoint calls, another for caching survey or instrument data, etc)

src/main/java/uk/gov/ons/ctp/response/collection/exercise/lib/common/redis/SessionResolver.java
```diff
@@ -0,0 +1,22 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(SessionResolver.class);
+
+  private static final String REDIS_KEY_PREFIX = "frontstage:collection-exercise-by-survey-id:";
+
+  @Autowired private RedisUtil<Object> redisUtil;
+
+  public void removeActiveSession(String surveyId) {
```
insacuri, line 17: This function name also seems wrong.  Theres no session involved here, all this key is doing is just caching the result of an endpoint call.  If we end up having a function per key you'd want to delete, this could maybe be 'deleteFrontstageExerciseByIdKey' or something similarly descriptive.

Also a comment on this saying which endpoint call was originally called could be useful for context?  But I won't block the PR if you think it's not worth it.

insacuri (changes requested):

insacuri:
Some slightly improved 'how to test' instructions on this card could be helpful too for the testers.  Just letting them know to log in a minute or two before the golive time, keep refreshing the page then within a minute of the go live time to notice that it should now appear.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/EventService.java
```diff
@@ -397,6 +400,8 @@ public void processEvents() {
                   collectionExerciseService.transitionCollectionExercise(
                       event.getCollectionExercise(),
                       CollectionExerciseDTO.CollectionExerciseEvent.GO_LIVE);
+                  sessionResolver.removeActiveSession(
+                      String.valueOf(event.getCollectionExercise().getSurveyId()));
```
insacuri, line 404: What happens in the error state for this by the way?  I think as it stands, it would stop everything because it would throw an exception, bubble up and just raise a 500 for the cronjob to ignore.

I don't think deleting the key should be a showstopper as it naturally will go away in a maximum of 10 minutes.  You should catch whatever exception it throws, log an error and carry on like all the other statements in this function generally do.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/lib/common/redis/EndpointCacheDataDeleter.java
```diff
@@ -0,0 +1,29 @@
...
+
+  @Autowired private RedisUtil<Object> redisUtil;
+
+  public void deleteFrontstageExerciseByIdKey(String surveyId) {
+    String key = REDIS_KEY_PREFIX + surveyId;
+    try {
+      Boolean cacheInvalidated = redisUtil.deleteValue(key);
+      log.with("key", key).with("success", cacheInvalidated).info("Cache invalidated with key");
+    } catch (Exception e) {
+      log.with("key", key).error("Unable to remove key from redis", e);
```
insacuri, line 26: Could you put a comment here saying something like 'Not bubbling up the exception intentionally as a failed key removal isn't fatal and it has a ttl so it will expire naturally'

insacuri (approved):
LGTM after the last comment I had around adding a comment in the error handling section

PR #312, opened 2023-06-06
RAS-728 Ready for live button not appearing

LJBabbage:
# What and why?
This PR removes the collection instrument id from the message and logging. 
We now update multiple collection instruments at a time which makes this largely redundant. Calling the endpoint multiple times as we have been doing recently is inefficient as it is forcing it to do state checks each time when it should be done just one

Additionally having the uuid of another services object is a bit unusual, specially as it is only used for the logs, that logging should be captured by the CI service, and this one should log what called it and what it affects in its service

N.B This service needs further work as it receives a call from CI and then makes a call back to CI, it shouldn't be doing that and be provided the details it needs on the first call, however as this is a bug fix for something else, it should be captured on another card 

# How to test?
Test with https://github.com/ONSdigital/response-operations-ui/pull/868 and https://github.com/ONSdigital/ras-collection-instrument/pull/307 (description on collection instrument)
# Trello

### DIFF
--- a/_infra/helm/collection-exercise/Chart.yaml
+++ b/_infra/helm/collection-exercise/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 13.0.12
+version: 13.0.13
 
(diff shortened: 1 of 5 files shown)


SteveScorfield (approved):
Looks good to me

PR #318, opened 2023-09-07
RAS-872: Extend EQ launch claims to include sds_dataset_id if required for the form_type

SteveScorfield:
# What and why?
The SDS ID is required to be passed to eQ when an SDS is present and the form type(s) in question match that supplied by SDS. This PR achieves this by grabbing the SDS attributes from the database and passing this to Fronstage for further processing.

# How to test?
1. Deploy this PR to your local environment
2. Add a CE in rOps and then add an SDS that matches the CE - this can be done by either directly adding it to the DB or adding it to the pubsub subscription 'supplementary-data-service-subscription-[YOUR-ENV]'.
3. Hit the endpoint: '/collectionexercises/[EXERCISE-ID]' of the collection exercise service
4. The SDS values should be present within the 'supplementaryDatasetJson' node of the returned JSON

Note: You can also test this when an SDS is not present. The 'supplementaryDatasetJson' will be null.

# Jira
https://jira.ons.gov.uk/browse/RAS-872

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
@@ -96,7 +96,6 @@ public class CollectionExercise {
   @Column(name = "eq_version")
   private String eqVersion;
 
-  //  @OneToOne(optional = true) is true by default
   @OneToOne(mappedBy = "collectionExercise")
   @JsonManagedReference
   private SupplementaryDatasetEntity supplementaryDatasetEntity;
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
@@ -1,5 +1,6 @@
 package uk.gov.ons.ctp.response.collection.exercise.domain;
 
+import java.io.Serializable;
 import java.sql.Timestamp;
 import java.util.UUID;
 import javax.persistence.Column;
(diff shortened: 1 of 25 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -66,4 +66,12 @@ private SupplementaryDatasetEntity createSupplementaryDatasetEntity(
           e);
     }
   }
+
+  public SupplementaryDatasetEntity findSupplementaryDataset(int exercise_pk) {
```
JacobHarrisonONS, line 70: rename to exercise_pk to exercisePk

src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
```diff
@@ -93,4 +95,9 @@ public class CollectionExercise {
 
   @Column(name = "eq_version")
   private String eqVersion;
+
+  //  @OneToOne(optional = true) is true by default
```
JacobHarrisonONS, line 99: delete commented code

JacobHarrisonONS (changes requested):

src/main/java/uk/gov/ons/ctp/response/collection/exercise/domain/CollectionExercise.java
SteveScorfield: Done

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
SteveScorfield: Done

PR #320, opened 2023-10-10
Add sample size and links to the Collex DTO

mfcrocker:
# What and why?
Currently there are a number of endpoints that return a CollectionExerciseDTO object (which then gets converted into json).  This object has a bunch of useful information about a collection exercise, including the action events, but it's missing any information about the sample summary and the sample size.  This matters because we have to make repeated calls to the exercise service to get this information in our other services.

# How to test?
Do any request that returns a CollectionExerciseDTO - I would suggest GET /collectionexercises/{collex_id}/survey/{survey_id}

# Jira
RAS-302

### DIFF
--- a/_infra/helm/collection-exercise/Chart.yaml
+++ b/_infra/helm/collection-exercise/Chart.yaml
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 13.0.19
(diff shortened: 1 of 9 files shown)


JacobHarrisonONS (approved):
Curled an endpoint that returns collectionExerciseDTO and the sampleSize and list of sample ids is present (output is on jira ticket). Run acceptance tests against branch and they were fine. services function correctly.

PR #325, opened 2023-11-20
RAS-930 Reduce collection exercise logging level to info

arroyoAle:
# What and why?
Updating the logging level from debug to info
# How to test?
deploy to you environment and run the acceptance tests, make sure that no `debug` level logs appear in the collection exercise logs
# Jira
https://jira.ons.gov.uk/browse/RAS-930

### DIFF
--- a/_infra/helm/collection-exercise/Chart.yaml
+++ b/_infra/helm/collection-exercise/Chart.yaml
@@ -14,9 +14,9 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 13.0.23
+version: 13.0.24
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 13.0.23
+appVersion: 13.0.24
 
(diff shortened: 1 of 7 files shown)


LJBabbage (changes requested):
There are a couple of log line in there that need upping

"Event is not actionable, setting to COMPLETED state", "Problem loading fixture {} reason {}"

A couple I am on the fence with i.e "Deleting collection exercise event id, event tag". I am not bothered about getting DTOs or hitting end points etc, but deleting things I am a bit more interested to see them in the logs, at a stretch an important creation as well.

arroyoAle:
> There are a couple of log line in there that need upping
> 
> "Event is not actionable, setting to COMPLETED state", "Problem loading fixture {} reason {}"
> 
> A couple I am on the fence with i.e "Deleting collection exercise event id, event tag". I am not bothered about getting DTOs or hitting end points etc, but deleting things I am a bit more interested to see them in the logs, at a stretch an important creation as well.

I've updated some of the debug logs, mostly to info, but the ones which are around problems were elevated to warn (they're also in the test directory)

LJBabbage (approved):

mfcrocker (approved):

PR #329, opened 2023-12-06
Revert "change debug to info"

anwilkie:
Reverts ONSdigital/rm-collection-exercise-service#326

This PR needed to be reverted because I only changed the debugging levels for the purpose of testing things out in preprod, but we don't want these logging statements showing up in prod.

To test, run acceptance tests and check the logs for the pod. They shouldn't show up (use CTRL+F and type the logging messages to check) becuase they've gone back to being debug messages.


LJBabbage (approved):

matthew-robinson-ons (approved):
revert changes look good

PR #333, opened 2024-02-22
Improve exception handling and logging for SDS

matthew-robinson-ons:
# What and why?
This PR improves the exception handling and logging of the SDS consumer. It introduces a new resource not found exception for when the SDS message doesn't return a valid collection exercise. It also improves the logging so the period and survey id of the message is logged and an explicit message is displayed when the collection exercise cannot be found. 
# How to test?
Publish a message to your topic:
```
gcloud pubsub topics publish projects/ras-rm-dev/topics/supplementary-data-service-<YOUR_NAMESPACE> --message='{
      "survey_id": "009",
      "period_id": "2025",
      "form_types": [
         "0001"
        ],
      "title": "MBS",
      "sds_published_at": "2023-08-24T11:55:00Z",
      "total_reporting_units": 10,
      "schema_version": "v1.0.0",
      "sds_dataset_version": 3,
      "filename": "815e0310-91a9-48d4-82a6-e827966421d2.json",
      "dataset_id": "815e0310-91a9-48d4-82a6-e827966421d2"
      }'
```
Check the log output
Deploy this PR to your environment. 
Publish a message to your topic and check the logs for the updated log information. Should work for null values as well.
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&view=detail&selectedIssue=RAS-1083

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
@@ -28,7 +28,16 @@ public class SupplementaryDatasetService {
     CollectionExercise collectionExercise =
         collectionExerciseService.findCollectionExercise(
             supplementaryDatasetDTO.getSurveyId(), supplementaryDatasetDTO.getPeriodId());
-
+    if (collectionExercise == null) {
+      log.error(
+              "Failed to find collection exercise for supplementary dataset. "
+                      + "survey_id: {}, period_id: {}",
+              supplementaryDatasetDTO.getSurveyId(),
+              supplementaryDatasetDTO.getPeriodId());
+      throw new CTPException(
+              CTPException.Fault.RESOURCE_NOT_FOUND,
+              "Failed to find collection exercise for supplementary dataset");
+    }
     try {
       if (existsByExerciseFK(collectionExercise.getExercisePK())) {
         log.info(
@@ -42,16 +51,8 @@ public class SupplementaryDatasetService {
       supplementaryDatasetRepository.save(supplementaryDatasetEntity);
       log.info("Successfully saved the supplementary dataset to the database");
     } catch (Exception e) {
-      if (collectionExercise == null) {
-        log.error(
-            "Failed to find collection exercise for supplementary dataset. "
-                + "SurveyId: {}, PeriodId: {}",
-            supplementaryDatasetDTO.getSurveyId(),
-            supplementaryDatasetDTO.getPeriodId());
-        throw new CTPException(
-            CTPException.Fault.RESOURCE_NOT_FOUND,
-            "Failed to find collection exercise for supplementary dataset");
-      }
+      throw new CTPException(
+              CTPException.Fault.SYSTEM_ERROR, "Something went wrong adding dataset", e);
     }
   }
 
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
@@ -42,8 +42,13 @@ public class SupplementaryDatasetService {
       supplementaryDatasetRepository.save(supplementaryDatasetEntity);
       log.info("Successfully saved the supplementary dataset to the database");
     } catch (Exception e) {
-      throw new CTPException(
-          CTPException.Fault.SYSTEM_ERROR, "Something went wrong adding dataset", e);
+      if (collectionExercise == null) {
+        log.error("Failed to find collection exercise for supplementary dataset. " +
+                "SurveyId: {}, PeriodId: {}", supplementaryDatasetDTO.getSurveyId(), supplementaryDatasetDTO.getPeriodId());
+        throw new CTPException(
+                CTPException.Fault.RESOURCE_NOT_FOUND,
+                "Failed to find collection exercise for supplementary dataset");
+      }
     }
   }
 
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
@@ -43,11 +43,14 @@ public class SupplementaryDatasetService {
       log.info("Successfully saved the supplementary dataset to the database");
     } catch (Exception e) {
       if (collectionExercise == null) {
-        log.error("Failed to find collection exercise for supplementary dataset. " +
-                "SurveyId: {}, PeriodId: {}", supplementaryDatasetDTO.getSurveyId(), supplementaryDatasetDTO.getPeriodId());
+        log.error(
+            "Failed to find collection exercise for supplementary dataset. "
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -42,8 +42,16 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
       supplementaryDatasetRepository.save(supplementaryDatasetEntity);
       log.info("Successfully saved the supplementary dataset to the database");
     } catch (Exception e) {
-      throw new CTPException(
-          CTPException.Fault.SYSTEM_ERROR, "Something went wrong adding dataset", e);
+      if (collectionExercise == null) {
+        log.error(
+            "Failed to find collection exercise for supplementary dataset. "
+                + "SurveyId: {}, PeriodId: {}",
```
pricem14pc, line 48: Although the DTO fields for this object are `surveyId` and `periodId` (note the lowercasing), the values passed in the message are `survey_id` and `period_id`. As such we should keep the error message consistent and use those field names. It will avoid confusion and aid log searching.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -42,8 +42,16 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
...
+            supplementaryDatasetDTO.getSurveyId(),
+            supplementaryDatasetDTO.getPeriodId());
+        throw new CTPException(
```
pricem14pc, line 51: Have we lost the original handling for the other (`SYSTEM_ERROR`) exceptions that may occur in this try block (there are also SDS database activity happening that might go wrong)?

It might be worth leaving this try block as it was, and throw a completely separate `RESOURCE_NOT_FOUND` exception straight after we've tried to retrieve the collection exercise (if it's null). That would keep it explicit and a bit cleaner.

pricem14pc (changes requested):

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
matthew-robinson-ons: I have update the log messages to use the values `survey_id` and `period_id`

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
matthew-robinson-ons: Yes I have reintroduced the original try block and moved the resource not found error to get called after the collection exercise call is made.

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetServiceTest.java
```diff
@@ -70,6 +70,23 @@ public void testFailedToFindCollectionExerciseForSupplementaryDataset() throws C
     supplementaryDatasetService.addSupplementaryDatasetEntity(supplementaryDatasetDTO);
   }
 
+  @Test
+  public void testFailedToFindCollectionExerciseForSupplementaryDatasetResourceNotFoundException() {
```
pricem14pc, line 74:
```suggestion
  public void testFailedToFindCollectionExerciseThrowsResourceNotFoundException() {
```

pricem14pc:
Thinking about the original `testFailedToFindCollectionExerciseForSupplementaryDataset` this is probably redundant now.

src/test/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetServiceTest.java
```diff
@@ -1,8 +1,8 @@
 package uk.gov.ons.ctp.response.collection.exercise.service;
 
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.junit.Assert.*;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.*;
```
pricem14pc, line 5: Have we considered the implications of changing this to use wildcarded Mocito package imports here?

matthew-robinson-ons: Looked into this and have updated the junit wildcard import to be explicit as there are only two being imported. Have left the mockito wild carded for now to keep the import list concise.

SteveScorfield (approved):
This has been tested and reviewed. Code changes work as expected. PR approved.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -28,7 +28,17 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
     CollectionExercise collectionExercise =
         collectionExerciseService.findCollectionExercise(
             supplementaryDatasetDTO.getSurveyId(), supplementaryDatasetDTO.getPeriodId());
-
+    if (collectionExercise == null) {
+      log.error(
```
pricem14pc, line 32: As exceptions are logged to stderr by default, we probably don't need this line as it will simply duplicate the throw message below. Now we've added the `survey_id` and `period_id` into the exception we should have all the info we need to alert on.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -28,7 +28,12 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
     CollectionExercise collectionExercise =
         collectionExerciseService.findCollectionExercise(
             supplementaryDatasetDTO.getSurveyId(), supplementaryDatasetDTO.getPeriodId());
-
+    if (collectionExercise == null) {
+      throw new CTPException(
+          CTPException.Fault.RESOURCE_NOT_FOUND,
+          "Failed to find collection exercise for supplementary dataset "
+              + "survey_id: {}, period_id: {}");
```
pricem14pc, line 35: The CTPException uses `String.format` which requires `%s` as the placeholders for interpolation rather than `{}`. We should include the values of the DTO under test in the assert of the message to ensure the `period_id` and `survey_id` are correctly included in the exception message.

pricem14pc (changes requested):

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
matthew-robinson-ons: Have updated the code to use string.format, also made the tests more explicit by using the survey and period id strings to improve the testing.

pricem14pc (approved):

PR #337, opened 2024-06-04
Remove EQ version

mfcrocker:
# What and why?
EQ v2 has been dead for some time, this removes the concept of a particular version.

# How to test?

# Jira
RAS-1144


LJBabbage (changes requested):
I don't think we can reformat and do the change at the same time +1,092 -1,017 makes it hard to tell what has changed

mfcrocker:
> I don't think we can reformat and do the change at the same time +1,092 -1,017 makes it hard to tell what has changed

As discussed I've tried and can't provide a nicer diff. CEEndpoint removes `patchCollectionExerciseEqVersion` and the CEEndpointUnitTests removes `testPatchCollectionExerciseEQVersion` - there shouldn't be any other non-formatting changes in those files

LJBabbage:
> > I don't think we can reformat and do the change at the same time +1,092 -1,017 makes it hard to tell what has changed
> 
> As discussed I've tried and can't provide a nicer diff. CEEndpoint removes `patchCollectionExerciseEqVersion` and the CEEndpointUnitTests removes `testPatchCollectionExerciseEQVersion` - there shouldn't be any other non-formatting changes in those files

I can't seem to see patchCollectionExerciseEqVersion at all https://github.com/search?q=repo%3AONSdigital%2Frm-collection-exercise-service%20patchCollectionExerciseEqVersion&type=code. And tried EqVersion in my ide

LJBabbage (approved):

arroyoAle (approved):
Looks good and tested and works as expected

PR #338, opened 2024-06-17
RAS-917 SupplementaryDatasetService dataset call via existsByExerciseFK not required

arroyoAle:
# What and why?
Changes how we check if an SDS in already linked to a collection exercise since the collection exercise already has the liked sds in scope through the ORM instead of querying the database every time.
# How to test?
- Deploy to your env
- Run the acceptance tests
- publish a message to your sds topic
```
{
      "survey_id": "139",
      "period_id": "1912",
      "form_types": [
         "0001"
        ],
      "title": "Quarterly Business Survey | QBS | 139 | 1912",
      "sds_published_at": "2023-08-24T11:55:00Z",
      "total_reporting_units": 1,
      "schema_version": "v1.0.0",
      "sds_dataset_version": 2,
      "filename": "eccb61ad-4eb3-4714-8afd-8c8426960b43.json",
      "dataset_id": "eccb61ad-4eb3-4714-8afd-8c8426960b43"
}
```
- make sure that in the logs there is no linked collection exercise found and that there is no log saying that the sds has been removed
- check an sds entry has been added to the database
- publish the same message again (you can edit the sds id to make sure that the old one has been removed and a new one added)
- Make sure the above logs are now present and the new sds entry is in the database`
# Jira
[RAS-917](https://jira.ons.gov.uk/browse/RAS-917)

### DIFF
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
@@ -39,7 +39,7 @@ public class SupplementaryDatasetService {
     try {
       if (collectionExercise.getSupplementaryDatasetEntity() != null) {
         log.info(
-            "Supplementary dataset linked to collection exercise {} found.",
+            "Supplementary dataset linked to collection exercise with PK {} found.",
             collectionExercise.getExercisePK());
         supplementaryDatasetRepository.deleteByExerciseFK(collectionExercise.getExercisePK());
         log.info("Supplementary dataset has been removed.");
--- a/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
+++ b/src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
@@ -39,7 +39,7 @@ public class SupplementaryDatasetService {
     try {
       if (collectionExercise.getSupplementaryDatasetEntity() != null) {
         log.info(
-            "Supplementary dataset linked to {} collection exercise found.",
+            "Supplementary dataset linked to the {} collection exercise found.",
             collectionExercise.getId());
         supplementaryDatasetRepository.deleteByExerciseFK(collectionExercise.getExercisePK());
         log.info("Supplementary dataset has been removed.");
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -37,9 +37,9 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
           supplementaryDatasetDTO.getPeriodId());
     }
     try {
-      if (existsByExerciseFK(collectionExercise.getExercisePK())) {
+      if (collectionExercise.getSupplementaryDatasetEntity() != null) {
```
LJBabbage, line 40: !=null do you need that?

arroyoAle: Yeah, if removed it just gets the `SupplementaryDatasetEntity` object and is expecting a `boolean` which java is not happy to deal with in an if

pricem14pc: Conceptually it's valid that the collection exercise does not have a mapped supplementary dataset entity in the database. As such, if we have mapped these objects so that its null if one doesn't exist, then this condition would be correct.

pricem14pc:
There's an `@SDS` acceptance test that can be run against this build too.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
```diff
@@ -37,9 +37,9 @@ public void addSupplementaryDatasetEntity(SupplementaryDatasetDTO supplementaryD
...
         log.info(
-            "Supplementary dataset with exerciseFk {} has been found.",
+            "Supplementary dataset linked to collection exercise with PK {} found.",
```
pricem14pc, line 42: Is the exercise UUID in scope? This would be preferable in the log entry to the PK.

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
pricem14pc: Oh I see sorry, you're both having a Python discussion!

src/main/java/uk/gov/ons/ctp/response/collection/exercise/service/SupplementaryDatasetService.java
arroyoAle: I've updated the log to have the collex uuid instead

pricem14pc (approved):

LJBabbage (approved):

pricem14pc:
Moved to verified branch https://github.com/ONSdigital/rm-collection-exercise-service/pull/340

PR #341, opened 2024-07-16
RAS-1184 Seft-ci-ingestion cloud function: Introduce HTTP client to retrieve collex

arroyoAle:
# What and why?
Changing collection exercise service to a NodePort instead of a ClusterIP to allow access from outside the kubernetes environment to give access to the cloud function
# How to test?
Follow test for https://github.com/ONSdigital/ras-rm-cloud-functions/pull/52
# Jira
https://jira.ons.gov.uk/browse/RAS-1184

### DIFF
--- a/_infra/helm/collection-exercise/templates/loadbalancer.yml
+++ b/_infra/helm/collection-exercise/templates/loadbalancer.yml
@@ -4,11 +4,11 @@ metadata:
   name: collection-exercise-lb
   annotations:
     networking.gke.io/load-balancer-type: "Internal"
-    networking.gke.io/load-balancer-ip-addresses: "collection-exercise-ip"
+    networking.gke.io/load-balancer-ip-addresses: {{ .Values.loadBalancer.ipAddress }}
 spec:
   type: LoadBalancer
   externalTrafficPolicy: Cluster
-  loadBalancerIP: "collection-exercise-ip"
+  loadBalancerIP: {{ .Values.loadBalancer.ipAddress }}
   selector:
     app: collection-exercise
   ports:
--- a/_infra/helm/collection-exercise/templates/loadbalancer.yml
+++ b/_infra/helm/collection-exercise/templates/loadbalancer.yml
@@ -4,11 +4,11 @@ metadata:
   name: collection-exercise-lb
   annotations:
     networking.gke.io/load-balancer-type: "Internal"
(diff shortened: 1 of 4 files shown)


_infra/helm/collection-exercise/values.yaml
```diff
@@ -72,4 +72,7 @@ gcp:
   collectionExerciseEndTopic: collection-exercise-end
   collectionExerciseEventStatusUpdateSubscription: collection-exercise-event-update
   supplementaryDataServiceTopic: supplementary-data-service-topic
-  supplementaryDataServiceSubscription: supplementary-data-service-subscription
\ No newline at end of file
+  supplementaryDataServiceSubscription: supplementary-data-service-subscription
+
+loadBalancer:
+  ipAddress: "10.110.128.12"
```
pricem14pc, line 78: Asd this is very specific internal config per env we shouldn't default this to anything here. It should only ever be provided by config.

_infra/helm/collection-exercise/templates/loadbalancer.yml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: collection-exercise-lb
+  annotations:
+    networking.gke.io/load-balancer-type: "Internal"
+    networking.gke.io/load-balancer-ip-addresses: {{ .Values.loadBalancer.ipAddress }}
```
pricem14pc, line 7: Worth noting that this is favoured over `spec.loadBalancerIP` for GKE > 1.2X(?). They're not both required but we will be running different versions of (GKE some < 1.2X ?) so until we move completely off the deprecated version I guess we can have both settings in place (even if it looks confusing)

SteveScorfield (approved):
Code looks fine. PR as a whole test in https://github.com/ONSdigital/ras-rm-cloud-functions/pull/52

anwilkie (approved):

_infra/helm/collection-exercise/values.yaml
arroyoAle: This IP is specifically only for dev and sandbox and will be replaced with a placeholder string once we upgrade to GKE 1.29

pricem14pc: Still not convinced this key is needed here at all.

pricem14pc (approved):

PR #344, opened 2024-09-19
update springboot to 2.7

anwilkie:
# What and why?
Updating springboot version to 2.7

# How to test?

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-1282)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
     <description>CollectionExerciseService</description>
 
     <properties>
-        <springboot.version>2.6.6</springboot.version>
+        <springboot.version>2.7.0</springboot.version>
         <spring.cloud.gcp.version>1.2.8.RELEASE</spring.cloud.gcp.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -10,7 +10,7 @@
     <description>CollectionExerciseService</description>
 
     <properties>
-        <springboot.version>2.6.6</springboot.version>
+        <springboot.version>2.7.0</springboot.version>
```
LJBabbage, line 13: There are later semantic bug fixes associated with 2.7

LJBabbage (changes requested):

SteveScorfield (approved):
Fully tested and works fine. Dependency upgrades look fine. Approved.

LJBabbage (approved):

pom.xml
anwilkie: Updated the spring boot version to a later version.

PR #371, opened 2026-03-13
RAS-1806 resolve database migration errors in Postgres 15+

LJBabbage:
# What and why?
Trailing characters in 2 Changelog files are causing migration issues to Postgres 15+ . However I don't believe we need to fix them and they can be removed. 

**Persisted environments**
The table casetypedefault gets removed in the delta release-15 which cascaded to the sequences. So in persisted environments where old deltas have run and don't get run again we need to do nothing.

**New environments**
With regards to newly create environments, when the full changelog gets exercised, the sequences will just not be created, but as they both get removed later and aren't explicitly used inbetween deleting the delta shouldn't be an issue

# How to test?
Use the pipline "Build Environment CE PR-371" in cluster setup to use this PR. If you create an environment and then deploy this PR it won't work, as the deltas will get run in first and then won't be updated. **This must be a new environment and why the pipeline.** 

Now check in databasechangelog in CE and make sure 10.49.13-1 and 10.49.14-1 aren't present. Once you confirm that just check that CE is working as expected

# Jira


SteveScorfield (approved):
Happy with this approach. Tested fully and everything works as expected. Approved.

matthew-robinson-ons (approved):
The changes worked when testing in dev. Still unsure how this will work in persisted environments when we attempt to upgrade as you can't upgrade without removing or changing the deltas in dev so unsure the difference in the other envs but happy to approve if we think this approach will work.

PR #372, opened 2026-04-22
Update CE end topic for DFTS

JacobHarrisonONS:
# What and why?
To enable deletion of datasets within SDS, RAS will need to update their collection exercise end date topic to include the dataset GUID in the message and allow SDS to subscribe to this topic

# How to test?

# Jira


SteveScorfield (approved):
Looks good to me. Approved.

.java-version
```diff
@@ -1,2 +1 @@
-21
-
+24
```
arroyoAle: We should not be updating to java 24

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/dto/CollectionExerciseEndEventDTO.java
```diff
@@ -11,4 +11,6 @@
 @NoArgsConstructor(access = AccessLevel.PUBLIC)
 public class CollectionExerciseEndEventDTO {
   private UUID collectionExerciseId;
+  private String period;
+  private String surveyId;
```
arroyoAle: We need to add the sds id and the collection exercise end here as well and then add them when we create the message (and omit them when there is no sds attached to the collection exercise

arroyoAle (changes requested):

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/CollectionExerciseEndPublisher.java
```diff
@@ -16,14 +18,21 @@ public class CollectionExerciseEndPublisher {
...
-  public void sendCollectionExerciseEnd(UUID collectionExerciseId) {
+  public void sendCollectionExerciseEnd(UUID collectionExerciseId, String exercise_ref) {
+
+    SurveyDTO survey = this.surveyService.findSurveyByRef(exercise_ref);
+
     CollectionExerciseEndEventDTO collectionExerciseEndEventDTO =
         new CollectionExerciseEndEventDTO();
     collectionExerciseEndEventDTO.setCollectionExerciseId(collectionExerciseId);
+    collectionExerciseEndEventDTO.setPeriod(exercise_ref);
+    collectionExerciseEndEventDTO.setSurveyId(survey.getId());
```
arroyoAle: This should also be something like `survey.getRef()` since the id is just internal to us and sds won't be able to make sense of it

.java-version
JacobHarrisonONS: this is a mistake when i was setting jevn

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/dto/CollectionExerciseEndEventDTO.java
JacobHarrisonONS: added sds id and end date

src/main/java/uk/gov/ons/ctp/response/collection/exercise/message/CollectionExerciseEndPublisher.java
JacobHarrisonONS: changed to use ref

arroyoAle (approved):
Looks good, works as I expect and tested with and without sds data and sample and collection exercise work fine

